PR #11, opened 2026-02-17
Rationalize validation case result

SebastienBocquet:
Allow to create a validation point result from stochastic validation points.
So the same validation plots (predict vs true, error scatter matrix etc...) can be created either from a collection of validaiton points, or from a `DeterministicValidationCase`.


SebastienBocquet:
The how to 07_validation_case/plot_stochastic_validation should explain how nominal values can be extracted from the reference data, and how they can be passed to the `StochasticValidationPoint`.

PR #18, opened 2026-03-17
Integrate Open Hole Tan model

SebastienBocquet:


src/vimseo/problems/load_cases/open_hole.py
```diff
@@ -0,0 +1,25 @@
...
+from __future__ import annotations
+
+from dataclasses import dataclass
+
+from vimseo.core.load_case import LoadCase
+
+
+@dataclass
+class OHT(LoadCase):
+    """An Open Hole Compression load case."""
```
FlorentGRT: OHT => tension

src/vimseo/lib_vimseo/tan_lib.py
```diff
@@ -0,0 +1,258 @@
...
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this program; if not, write to the Free Software Foundation,
+# Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+
```
FlorentGRT, line 15: It is missing a doctsring for every function of this file

src/vimseo/lib_vimseo/tan_lib.py
```diff
@@ -0,0 +1,258 @@
...
+            [c**2, s**2, 2 * c * s],
+            [s**2, c**2, -2 * c * s],
+            [-c * s, c * s, c**2 - s**2],
+        ]),
+        10,
+    )
+
+
+def Effort_IeII(
+    R: float,  # rayon du trou
```
FlorentGRT: Many french comments to be translated

FlorentGRT (commented):
Here is a basic review, until I manage to actually try this model

src/vimseo/tools/bayes/bayes_analysis.py
```diff
@@ -117,17 +118,18 @@ class BayesSettings(BaseSettings):
...
-        default=empty,
+    x0s: ndarray = Field(
+        default=empty(0),
         description="The starting points of the algorithm. "
         "In practice a 1-D array of size the number "
         "of parameters of the model.",
     )
     n_mcmc: int = Field(
-        default=1_000, description="The number of steps of the mcmc analysis."
+        default=1_000, description="The number of steps of the mcmc analysis.yes"
```
SebastienBocquet: Fix typo yes

src/vimseo/problems/tan_oh/tan_oht.py
```diff
@@ -0,0 +1,176 @@
...
+    from collections.abc import Sequence
+
+LOGGER = logging.getLogger(__name__)
+
+DEFAULT_INPUT_DATA = {
+    "d0": atleast_1d(0.71),
+    "radius": atleast_1d(3.175),
+    "width": atleast_1d(32.0),
+    "length": atleast_1d(50.0),
+    "load": array([5000, 0, 0]),
```
FlorentGRT: new proposed default (helps the fields understanding + more realistic values) : 
load = [1000; 0;0]
lenght  = 80

src/vimseo/problems/tan_oh/tan_oht.py
```diff
@@ -0,0 +1,176 @@
...
+from vimseo.core.model_metadata import MetaDataNames
+from vimseo.core.model_settings import IntegratedModelSettings
+from vimseo.lib_vimseo.tan_lib import tan_model
+
+if TYPE_CHECKING:
+    from collections.abc import Mapping
+    from collections.abc import Sequence
+
+LOGGER = logging.getLogger(__name__)
+
```
FlorentGRT, line 59: I checked the VTK field, it seems nice

src/vimseo/problems/tan_oh/tan_oht.py
```diff
@@ -0,0 +1,176 @@
...
+DEFAULT_INPUT_DATA = {
+    "d0": atleast_1d(0.71),
+    "radius": atleast_1d(3.175),
+    "width": atleast_1d(32.0),
+    "length": atleast_1d(50.0),
+    "load": array([5000, 0, 0]),
+    "n_x": atleast_1d(100),
```
FlorentGRT: add grammar for n_x and and n_y, especially minium bounds ?

FlorentGRT (commented):
my review is finished

src/vimseo/problems/load_cases/open_hole.py
SebastienBocquet: Fixed

src/vimseo/lib_vimseo/tan_lib.py
SebastienBocquet: Ok I'll take care of this.

src/vimseo/lib_vimseo/tan_lib.py
SebastienBocquet: Ok I'll take care of this.

src/vimseo/problems/tan_oh/tan_oht.py
SebastienBocquet: Ok. I will move to the new values after I switch to a Pydantic grammar.

src/vimseo/problems/tan_oh/tan_oht.py
SebastienBocquet: I introduced a coarsening_factor, and a default constant for the nominal number of points along x and y.

src/vimseo/tools/bayes/bayes_analysis.py
SebastienBocquet: Done

src/vimseo/problems/tan_oh/tan_oht.py
SebastienBocquet: I did not implemente a json grammar. Currently this model is relying on a SimpleGrammar.

src/vimseo/tools/bayes/bayes_analysis.py
```diff
@@ -1,3 +1,18 @@
...
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this program; if not, write to the Free Software Foundation,
+# Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+
```
FlorentGRT, line 15: the credit are repeated in this file.

other files may have the same problem too ?

src/vimseo/problems/tan_oh/tan_oht.py
FlorentGRT: Now that it is clearer that we can use some custom pre_dataproc and post_dataproc for any given analysis, I am thinking it would be better to have the following practive:
- "mesh_size" as a clear basic input of any model
- "coarsening_factor" could be introduced by a custom pre_dataproc when needed, for exemple for a solution verification analysis

FlorentGRT:
2nd round of review

FlorentGRT (commented):
2nd review, actual submitting

src/vimseo/tools/bayes/bayes_analysis.py
SebastienBocquet: Should be good

FlorentGRT: no I still obseve repeated credits

docs/references/open_hole_tan_model/equations.md
```diff
@@ -0,0 +1,169 @@
...
+ Copyright 2021 IRT Saint Exupery, https://www.irt-saintexupery.com
+
+ This work is licensed under the Creative Commons Attribution-ShareAlike 4.0
+ International License. To view a copy of this license, visit
+ http://creativecommons.org/licenses/by-sa/4.0/ or send a letter to Creative
+ Commons, PO Box 1866, Mountain View, CA 94042, USA.
+-->
+
+<!--
+ Copyright 2021 IRT Saint Exupery, https://www.irt-saintexupery.com
```
FlorentGRT: credits doubled

docs/references/open_hole_tan_model/equations.md
FlorentGRT, line 1: missing "if" in condition "Nxx != Nyy" ?

FlorentGRT, line 1: wrong formatting of "T \alpha"  => unexpected space ?

FlorentGRT, line 1: "there correspond" => "there is"

FlorentGRT, line 1: layout of N1 and N2 : "1" and "2" should be indices

docs/references/open_hole_tan_model/index.md
```diff
@@ -0,0 +1,28 @@
...
+ Copyright 2021 IRT Saint Exupery, https://www.irt-saintexupery.com
+
+ This work is licensed under the Creative Commons Attribution-ShareAlike 4.0
+ International License. To view a copy of this license, visit
+ http://creativecommons.org/licenses/by-sa/4.0/ or send a letter to Creative
+ Commons, PO Box 1866, Mountain View, CA 94042, USA.
+-->
+
+<!--
+ Copyright 2021 IRT Saint Exupery, https://www.irt-saintexupery.com
```
FlorentGRT: credits doubled

docs/references/materials/ta6v.md
FlorentGRT, line 10: TA6V is irrelevant for Tan

docs/references/open_hole_tan_model/validity.md
FlorentGRT, line 1: `However, at the edge of the hole, this approximation leads to infinite stresses that are not physically realistic. Therefore, it is necessary to consider a region of validity for Tan’s model that excludes a certain area around the hole. To do this, we increase the radius of the hole by a distance d0`

Several assertions are false here. Alternative proposition : 

`However, when damage occurs (matrix cracking, and possibly later some fibers failure), there are some local stress redistributions and the considered linear elastic law becomes invalid locally. Engineers use the "point-stress" method, as a correction of basic linear-elastic results, in order to estimate the occurence of substancial damage propagation (fibre failure). This method consist in observing the stress value (computed from a fully elastic structure) at a distance 'd_0' from the hole ("point-stress distance"), instead of the direct vicinity of the hole. Despite some physical meaningfulness of 'd_0', its value is actually identified from tests and it is highly sensitive to many parameters (material ply properties, layup, size of the hole, etc).`

FlorentGRT, line 1: "In the previous figure, the stress concentration zone is clearly visible, where the elastic model is no longer valid."

False. Alternative proposition : 
"The previous figure, the stripped area is where the computed stress is ignored, for estimation of damage occurence. Note that for a critical loading, it also match approximately where the elastic model is no longer valid."

src/vimseo/problems/tan_oh/tan_oht.py
```diff
@@ -0,0 +1,502 @@
...
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this program; if not, write to the Free Software Foundation,
+# Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+
+# Copyright 2021 IRT Saint Exupery, https://www.irt-saintexupery.com
```
FlorentGRT: credits doubled

src/vimseo/problems/tan_oh/tan_oht.py
```diff
@@ -0,0 +1,502 @@
...
+class TanRun_Tension(BaseComponent):
+    """An Open Hole Tension model based on Tan theory (#open-hole-plate-model-tan-model).
+
+    No grammar is defined for the material (and thus no bounds).
+    """
+
+    USE_JOB_DIRECTORY = True
+
+    auto_detect_grammar_files = False
+    default_grammar_type = "JSONGrammar"
```
FlorentGRT, line 118: Is it not a SimpleGrammar here ?

src/vimseo/problems/tan_oh/tan_oht.py
```diff
@@ -0,0 +1,502 @@
...
+        })
+
+    def _run(self, input_data):
+
+        length = input_data["length"][0]
+        width = input_data["width"][0]
+        thickness = input_data["thickness"][0]
+        input_data["d0"][0]
+        radius = input_data["radius"][0]
+        coarsening_factor = input_data["coarsening_factor"][0]
```
FlorentGRT: I feel like we should not have "coarsening_factor" as input of any model, but "mesh_size" typically. For this specific model, "grid_size" or "grid_step_size" would be better to avoid confusion with other models actual mesh refinement.
It should be the responsability of a workflow/tool/analysis (like SolutionVerification) to performs inputs conditionning like mesh_size=>coarsening_factor.
@SebastienBocquet this issue needs to be discussed orally

src/vimseo/problems/tan_oh/tan_oht.py
```diff
@@ -0,0 +1,502 @@
...
+        output_data["dy"] = atleast_1d(dy)
+
+        return output_data
+
+
+class PostFieldExtraction(BaseComponent):
+    """A post-processor to extract data from a field."""
+
+    auto_detect_grammar_files = False
+    default_grammar_type = "JSONGrammar"
```
FlorentGRT, line 223: Idem, SimpleGrammar ?

src/vimseo/problems/tan_oh/tan_oht.py
```diff
@@ -0,0 +1,502 @@
...
+                    load_case=LoadCaseFactory().create(load_case_name),
+                    material_grammar_file=MATERIAL_GRAMMAR_FILE,
+                    material=material,
+                ),
+                PostFieldExtraction(
+                    load_case=LoadCaseFactory().create(load_case_name),
+                    fields_from_file=self.FIELDS_FROM_FILE,
+                    material_grammar_file=MATERIAL_GRAMMAR_FILE,
+                    material=material,
+                ),
```
FlorentGRT, line 359: There is only one load_case so the factory is not relevent ? There should be a second load_case for a custom loading ?

src/vimseo/problems/tan_oh/tan_oht.py
```diff
@@ -0,0 +1,502 @@
...
+            "dx": atleast_1d(0.0),
+            "dy": atleast_1d(0.0),
+        })
+
+    def _run(self, input_data):
+
+        length = input_data["length"][0]
+        width = input_data["width"][0]
+        thickness = input_data["thickness"][0]
+        input_data["d0"][0]
```
FlorentGRT: weird, no value affected 
either use d0 value or do `_ = ...`  or remove this line ?

src/vimseo/problems/tan_oh/tan_oht.py
```diff
@@ -0,0 +1,502 @@
...
+        xx, yy = meshgrid(x, y, indexing="ij")  # shape (nx, ny)
+
+        # Flatten coordinates -> points array (nx*ny, 3)
+        points = column_stack([
+            xx.ravel(),
+            yy.ravel(),
+            zeros(n_x * n_y),  # z=0 for 2D
+        ])
+
+        d0_ = 0.0
```
FlorentGRT: why 0 value ? why not use the actual value of d0 ? => needs an explaination in the comment.

Variable declaration should be closer to its usage ?

src/vimseo/problems/tan_oh/tan_oht.py
```diff
@@ -0,0 +1,502 @@
...
+        # blank out the points falling inside the hole.
+        x_0 = xx - 0.5 * length
+        y_0 = yy - 0.5 * width
+        r = sqrt(x_0**2 + y_0**2)
+        theta = arctan2(y_0, x_0)
+
+        flux_n = tan_model_grid(
+            r.ravel(), theta.ravel(), load, c_strat, radius, width
+        ).reshape(n_x, n_y, 3)
+        flux_n[r < radius + d0_] = nan
```
FlorentGRT: Like in tan_validity_zone.png , on the output field plots it would be nice to add: 

- a circle line (black, not dotted) to represent the accurate position of the hole
- a circle line (blacj, dotted) to represent the point stress distance

src/vimseo/problems/tan_oh/tan_oht.py
```diff
@@ -0,0 +1,502 @@
...
+material = Material.from_json(MATERIAL_FILE)
+
+# Material property names driving the (elastic) membrane stiffness c_strat.
+STIFFNESS_PROPERTY_NAMES = ("E1", "E2", "G12", "nu12")
+
+total_thickness = len(DEFAULT_INPUT_DATA["stacking_sequence"]) * PLY_THICKNESS
+DEFAULT_INPUT_DATA["thickness"] = atleast_1d(total_thickness)
+
+
+def compute_c_strat(stacking_sequence, e1, e2, g12, nu12):
```
FlorentGRT: As a plot output of the model, it would be nice to have plot similar to the image below (from the curve histories you already have). The red vertical line should also be reproduced + an horizontal line mathcing the material critical stress.

This would push the Tan model towards more engineering use, instead of just the fields theoretical use case.


Image

src/vimseo/problems/tan_oh/tan_oht.py
```diff
@@ -0,0 +1,502 @@
...
+        material: Material | None = None,
+        check_subprocess: bool = False,
+        fields_from_file: Mapping[str, str] | None = None,
+    ):
+        super().__init__(
+            load_case=load_case,
+            material_grammar_file=material_grammar_file,
+            material=material,
+            check_subprocess=check_subprocess,
+        )
```
FlorentGRT, line 238: do **options ?

src/vimseo/problems/tan_oh/tan_oht.py
```diff
@@ -0,0 +1,502 @@
...
+        })
+        input_names.append("stacking_sequence")
+
+        for name in input_names:
+            self.input_grammar.required_names.add(name)
+
+        self._flux_components = ["sigma_xx", "sigma_yy", "sigma_xy", "Distance"]
+        self._line_name = "line_center"
+        for name in ["y", *self._flux_components]:
+            self.output_grammar.update_from_names([f"{self._line_name}_{name}"])
```
FlorentGRT: grammar definition should be clearer (explicitly readable) for such static outputs names => please define it from a basic list a static strings, instead of dynamically buildin them.

src/vimseo/problems/tan_oh/tan_oht.py
```diff
@@ -0,0 +1,502 @@
...
+        output_data["sigma_xx_r"] = atleast_1d(
+            thickness * tan_model(radius, 0.5 * pi, load, c_strat, radius, width)[0]
+        )
+        output_data["sigma_xx_d0"] = atleast_1d(
+            thickness
+            * tan_model(radius + d0, 0.5 * pi, load, c_strat, radius, width)[0]
+        )
+
+        # TODO: compute reserve factor based on strength criteria instead of just returning 1.0
+        output_data["reserve_factor"] = atleast_1d(1.0)
```
FlorentGRT, line 321: unfinished

failure_criterion = sigma_xx_d0 / sigma_fiber_failure           

sigma_fiber_failure  should be added to material inputs, and you can take sigma_fiber_failure  = 2300 MPa as a default material value

reserve_factor = 1 / failure_criterion

tests/core/test_model_composition.py
```diff
@@ -0,0 +1,149 @@
...
+from vimseo.api import create_model
+from vimseo.core.components.base_component import BaseComponent
+from vimseo.core.model_settings import IntegratedModelSettings
+
+
+def _make_post(base_model, extra_outputs=()):
+    """Build a post-processing component that depends on the base model grammars.
+
+    Mirrors the ``MyPost`` component of the ``plot_041_post_processing_patch``
+    gallery example. ``extra_outputs`` lets a test declare additional output
```
FlorentGRT: plot_041_post_processing_patch is from which library ?

tests/problems/test_tan_oht.py
```diff
@@ -0,0 +1,312 @@
...
+# this reference value to be deterministic and reproducible across BLAS/platforms.
+QUASI_ISOTROPIC_STACKING = np.array([0.0, 45.0, -45.0, 90.0, 90.0, -45.0, 45.0, 0.0])
+
+# A strongly orthotropic (0-dominant) stacking. Its roots s1 and s2 are well
+# separated (|s1 - s2| ~ 4.2), so the solution is well conditioned and
+# reproducible across platforms (see the module docstring of ``tan_lib``).
+ORTHOTROPIC_STACKING = np.array([0.0, 0.0, 90.0, 0.0, 0.0, 90.0, 0.0, 0.0])
+
+
+def _build_c_strat(stacking: np.ndarray) -> tuple[np.ndarray, float]:
```
FlorentGRT, line 54: It is quite redundant with `compute_c_strat()`
=> they should be factorised ?

docs/runnable_examples/05_solution_verification/plot_tan_open_hole_convergence.py
```diff
@@ -0,0 +1,430 @@
...
+The three-point Richardson extrapolation is fragile: its cross-validation returns
+``nan`` as soon as one grid triplet is not cleanly power-law convergent, which
+happens systematically for the sawtooth quantity but also for the smooth one on
+sampled data. The tool therefore also reports two Richardson-independent estimates
+of the converged value -- a least-squares power-law fit, whose order is fitted
+rather than assumed, and a model-free median of the finest grids -- which stay
+available in both cases. Their residual (fit) and band (median) quantify how
+reliable the estimate is.
+"""
+
```
FlorentGRT, line 44: @SebastienBocquet I did not manage to run the doc, however I need it to review this part of your branch.

Image

docs/runnable_examples/05_solution_verification/plot_tan_open_hole_convergence.py
```diff
@@ -0,0 +1,430 @@
...
+# %%
+from __future__ import annotations
```
FlorentGRT, line 46: TODO @FlorentGRT remaining work : 

- tox run -e doc
- cf plots generated by this file
- challenge the use of Tan model for a convergence study => use instead another model or be very clear on the mock meaning ? Challenge the metrics observed for convergence.

src/vimseo/material_lib/plane_orthotropic_ply_grammar.json
```diff
@@ -0,0 +1,107 @@
...
+      "description": "",
+      "items": {
+        "maximum": 1000000000000.0,
+        "minimum": -1000000000000.0,
+        "type": "number"
+      },
+      "title": "Nu12",
+      "type": "array"
+    },
+    "Xt": {
```
FlorentGRT, line 43: The failure stresses are defined here but are absent from the tan model usage.

src/vimseo/material_lib/composite.json
```diff
@@ -0,0 +1,228 @@
+{
+  "name": "composite",
```
FlorentGRT: The name of this material is too generic and will be prone to misunderstanding, especially within vimseo_composites.

We can create another name, more explicit, from these keywords : plane, ply, composite, orthotropic, elastic, max-stress.
Note that this material is very similar to EOMS, except it does not have out-of-plane properties. 

"PlaneOthotropicPly" ?

src/vimseo/material_lib/plane_orthotropic_ply.json
```diff
@@ -0,0 +1,228 @@
...
+        "misc": {},
+        "author": "",
+        "source": ""
+      },
+      "properties": [
+        {
+          "name": "E1",
+          "description": "",
+          "value": 135000000000.0,
+          "lower_bound": -1000000000000.0,
```
FlorentGRT: more realistic bounds may be considered for all inputs

FlorentGRT (commented):
2nd round of review done
missing doc review

requirements/test-python3.10.txt
FlorentGRT, line 1: a new update reqs seems to fetch a lot of new dependencies => update deps again ?

docs/runnable_examples/06_validation_point/plot_stochastic_validation_straightbeam.py
FlorentGRT, line 1: about 10 untracked result files are created with "tox -e run doc" => this seems to be an issue to be handled

SebastienBocquet:
Reverse precedence of model job dir on directpry_path:

Instead of:

```
  if save:
      directory_path = (
          self.archive_manager.job_directory
          if self.archive_manager.job_directory != ""
          else (Path(directory_path) if directory_path != "" else Path.cwd())
      )
      if not directory_path.exists():
          directory_path.mkdir(parents=True)
      LOGGER.info(f"Saving plots to {directory_path.absolute()}")
  else:
      directory_path = ""
```

```
if directory_path != "":
    directory_path = Path(directory_path)
elif self.archive_manager.job_directory != "":
    directory_path = Path(self.archive_manager.job_directory)
else:
    directory_path = Path.cwd()
if save:
      if not directory_path.exists():
          directory_path.mkdir(parents=True)
      LOGGER.info(f"Saving plots to {directory_path.absolute()}")
```

also protect the log on line 585 with an `if save`

src/vimseo/problems/tan_oh/tan_oht.py
SebastienBocquet: Ok, let's think about that in an issue.

For Tan model, do you want to switch to a json grammar to define bounds for non material inputs?

src/vimseo/tools/bayes/bayes_analysis.py
SebastienBocquet: Indeed. I will clean that after after integration of MR "coverage", because this MR also fixes the headers.

src/vimseo/problems/tan_oh/tan_oht.py
FlorentGRT: I let you choose the standing of this model : does it deserves a proper grammar or not

docs/references/open_hole_tan_model/equations.md
SebastienBocquet: Fixed

SebastienBocquet: Fixed

SebastienBocquet: Fixed

docs/runnable_examples/05_solution_verification/plot_tan_open_hole_convergence.py
SebastienBocquet: It seems that you have another mkdocs process blocking the serving

src/vimseo/material_lib/plane_orthotropic_ply_grammar.json
SebastienBocquet: Do we remove the Xt, Xc, Yt, Yc ?
Do we need them to compute the reserve factor (which is set to one currently)

src/vimseo/material_lib/composite.json
SebastienBocquet: Ok, I go for "PlaneOrthotropicPly", and "stacking_sequence" to "layup".

src/vimseo/material_lib/plane_orthotropic_ply.json
SebastienBocquet: I can go for a coefficient of variation of 10% around the nominal value.

requirements/test-python3.10.txt
SebastienBocquet: Yes, I update

docs/runnable_examples/06_validation_point/plot_stochastic_validation_straightbeam.py
SebastienBocquet: It is expected that some files are generated by the doc examples, and they are not tracked. If we track them, we would have to handle the diff at every doc build. It has the benefit of warning for any diff occuring in examples outputs, but complicates the change tracking. For me, the tests are responsible for tracking behaviour changes. I propose to keep the current way.

src/vimseo/material_lib/plane_orthotropic_ply.json
```diff
@@ -0,0 +1,228 @@
...
+            "scale": 1.0,
+            "rate": 1.0,
+            "lower_bound": -1000000000000.0,
+            "upper_bound": 1000000000000.0
+          }
+        },
+        {
+          "name": "E2",
+          "description": "",
+          "value": 10000000000.0,
```
FlorentGRT: wrong value for E2 : take 8E9 instead

src/vimseo/material_lib/plane_orthotropic_ply.json
```diff
@@ -0,0 +1,228 @@
...
+        },
+        "misc": {},
+        "author": "",
+        "source": ""
+      },
+      "properties": [
+        {
+          "name": "E1",
+          "description": "",
+          "value": 135000000000.0,
```
FlorentGRT: too much zeros in this file, prone to error : can you use powers of ten to represent big numbers ?

FlorentGRT: Usually we use another system of units (MPa instead of Pa) to avoid such large number. Please consider changing units.

In the meantime, the are two necessary steps to perform in order to avoid confusion : 
- units should be specified in the grammar description  
- The model description shoud warn that this uncommon units are used ?
- the grammar should be less permissive on lower bounds (no negative values in particular...) to avoid confusion with usual MPa values => see EOMS grammar in vimseo-abaqus

src/vimseo/material_lib/plane_orthotropic_ply_grammar.json
FlorentGRT: We need Xt and Xc to compute RF (sigma_computed/Xt => failure criterion => RF)

Yt and Yc may be use for more subtle damage initiation criterions, but let's remove them for now

docs/runnable_examples/06_validation_point/plot_stochastic_validation_straightbeam.py
FlorentGRT: expected untracked files/dirs should be declared in the git-ignore !?

If not, the git change utility is constantly pollutated by those files, and files needs to be manually deleted

docs/runnable_examples/05_solution_verification/plot_tan_open_hole_convergence.py
```diff
@@ -0,0 +1,430 @@
...
+    load_case,
+    model_options=IntegratedModelSettings(
+        directory_archive_root=EXAMPLE_RUNS_DIR / "archive/solution_verification",
+        directory_scratch_root=EXAMPLE_RUNS_DIR / "scratch/solution_verification",
+        cache_file_path=EXAMPLE_RUNS_DIR
+        / f"caches/solution_verification/{model_name}_{load_case}_cache.hdf",
+    ),
+)
+# The same job directory is reused across the runs of this study:
+model.archive_manager._accept_overwrite_job_dir = True
```
FlorentGRT: It is weird to have that line in the doc, especially for basic usage exemple

docs/runnable_examples/05_solution_verification/plot_tan_open_hole_convergence.py
```diff
@@ -0,0 +1,430 @@
+# Copyright 2021 IRT Saint Exupery, https://www.irt-saintexupery.com
+#
+# This work is licensed under a BSD 0-Clause License.
+#
+# Permission to use, copy, modify, and/or distribute this software
```
FlorentGRT, line 5: I had a hard time mapping this file to the content it generates in doc : I am not sure why the compiled doc section names must be (so) different from the dirs names in the sources here

docs/runnable_examples/05_solution_verification/plot_tan_open_hole_convergence.py
```diff
@@ -0,0 +1,430 @@
...
+- ``sigma_xx_r`` and ``sigma_xx_d0``: the native hole-edge stresses, evaluated
+  directly on the analytic solution and therefore exactly mesh-independent.
+
+The three-point Richardson extrapolation is fragile: its cross-validation returns
+``nan`` as soon as one grid triplet is not cleanly power-law convergent, which
+happens systematically for the sawtooth quantity but also for the smooth one on
+sampled data. The tool therefore also reports two Richardson-independent estimates
+of the converged value -- a least-squares power-law fit, whose order is fitted
+rather than assumed, and a model-free median of the finest grids -- which stay
+available in both cases. Their residual (fit) and band (median) quantify how
```
FlorentGRT: The end of the paragraoh is too specific and technical to be placed here as an introduction ?

docs/runnable_examples/05_solution_verification/plot_tan_open_hole_convergence.py
```diff
@@ -0,0 +1,430 @@
...
+
+- ``sigma_xx_probe``: ``sigma_xx`` bilinearly interpolated at a fixed point just
+  outside the hole. It converges smoothly as the grid is refined.
+- ``sigma_xx_peak``: the maximum of ``sigma_xx`` over the grid nodes. Because the
+  node that happens to fall closest to the stress concentration jumps around as
+  the grid changes, this quantity is *non-monotone* -- a sawtooth.
+- ``sigma_xx_r`` and ``sigma_xx_d0``: the native hole-edge stresses, evaluated
+  directly on the analytic solution and therefore exactly mesh-independent.
+
+The three-point Richardson extrapolation is fragile: its cross-validation returns
```
FlorentGRT: "RIchardson extrapolation" => add hyperlink reference to somewhere ?
Add context explanantion of the method ?

docs/runnable_examples/05_solution_verification/plot_tan_open_hole_convergence.py
```diff
@@ -0,0 +1,430 @@
...
+# OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES WHATSOEVER RESULTING
+# FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN ACTION OF CONTRACT,
+# NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF OR IN CONNECTION
+# WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+
+"""
+Solution verification of the analytic Tan open-hole model
+=========================================================
+
+Assess the discretization error of the :class:`TanOpenHole` model with the
```
FlorentGRT, line 21: This first paragraph is great. It can be even better : see below.

Following paragraphs are very technical purely specific to Tan model, which is not the interest of the reader here : the subject here is to perform a generic solution verification => the reader wants a minimal operational exemple, no complex explaination on the very specific (too complex ?) model used for this exemple. Also, I think the first paragraph can be even clearer, something towards the spirit of : "the objective here is to demonstrate a solution verification with VIMSEO, base on the `TanOpenHole` model as a use case here". 
The paragraphs 2-3-4 should be deleted/collapsed/referenced_elsewhere, an the barely useful content integrated in the first paragraph (see added content in bold) :

> Assess the discretization error of the :class:`TanOpenHole` model with the
>  ``DiscretizationSolutionVerification`` tool, and compare two field-derived
> quantities: one that converges smoothly **(sigma_xx_probe)** and one that exhibits a sawtooth
> behaviour **( sigma_xx_peak)** for which the Richardson extrapolation cannot be computed.
Others relevant metrics may also be integrated similarly in a minimal way.

docs/runnable_examples/05_solution_verification/plot_tan_open_hole_convergence.py
```diff
@@ -0,0 +1,430 @@
...
+``DiscretizationSolutionVerification`` tool, and compare two field-derived
```
FlorentGRT: The use of Tan model for solution verification is weird because it is not a proper mesh convergence issue like this tool is meant to tackle. Instead of technical explanations on the Tan model back room (which is precisely not the type of problem meant to be dealt with SolutionVerification), I think we should just say that "this model is used a use-case case, despite having a working principle different from what is usually prone to SolutionVerification", without further explanation (or further explanation hidden in a expand/collapse section.

docs/runnable_examples/05_solution_verification/plot_tan_open_hole_convergence.py
```diff
@@ -0,0 +1,430 @@
...
+
+# %%
+# We define a set of grid coarsening factors, from coarse to fine. The tool
+# requires strictly decreasing "element sizes"; here the coarsening factor plays
+# that role, as it is proportional to the grid spacing.
+coarsening_factors = [4.0, 3.0, 2.0, 1.5, 1.0, 0.5]
+
+
+# %%
+# The model writes the stress field to a VTK file. Converting the raw
```
FlorentGRT: Very technical and out of scope for minimal-useful Solution Verification

docs/runnable_examples/05_solution_verification/plot_tan_open_hole_convergence.py
```diff
@@ -0,0 +1,430 @@
...
+        combined.add_trace(trace, row=1, col=1)
+    for trace in fig_right.data:
+        # Avoid duplicating the shared legend entries in the right-hand panel.
+        trace.showlegend = False
+        combined.add_trace(trace, row=1, col=2)
+    combined.update_xaxes(title_text="coarsening_factor")
+    combined.update_yaxes(title_text=y_title, row=1, col=1)
+    return combined
+
+
```
FlorentGRT, line 106: I consider that "coarsening_factor" should not  be an input of a base model. However it is very relevant to define ad-hoc an envelopped model with a coarsening_factor new input.

docs/runnable_examples/05_solution_verification/plot_tan_open_hole_convergence.py
```diff
@@ -0,0 +1,430 @@
...
+# %%
+# The two field-derived quantities are not model outputs, so we compute them
```
FlorentGRT: I have a proposition on how to make it super clear by compartimentalisation of model setup (and technical comments) and Solution Verification usage.

On a dedicated separate file we could make a new model called `ModelToVerify`, by enveloping `TanOpenHole` . The new enveloped model gets "coarsening_factor" as an input, and expose minimal other inputs, with generic naming (not OH tainted). New useful outputs are also exposed (no field consideration), also with generic naming (something like "y_3" or "y_smooth"). In this file, we may keep all technical consideration regarding OH, fields, interpolation, ...

On this file, we may use `ModelToVerify` and focus on having the clrearest tutorial possible for solution verification, mostly/fully agnostic to the model specifics.

docs/runnable_examples/05_solution_verification/plot_tan_open_hole_convergence.py
```diff
@@ -0,0 +1,430 @@
...
+dataset = dataframe_to_dataset(
+    convergence_table.rename(
+        columns={
+            "coarsening_factor": "coarsening_factor{inputs}",
+            "sigma_xx_probe": "sigma_xx_probe{outputs}",
+            "sigma_xx_peak": "sigma_xx_peak{outputs}",
+        }
+    )
+)
+
```
FlorentGRT, line 167: model execution => ModelResult.from_data => field, "flux, sigma_xx, results  => probe_stresses , peak_stresses, model_results => DataFrame => dataframe_to_dataset => DiscretizationSolutionVerification

=> So much OH technicality + specific VIMSEO-GEMSEO formalisms + conditionning steps. I feel like it should be simpler to read as doc exemple + easier to use. Having `ModelToVerify` should make clearer + I guess we can shortcut the multi-step data-conditionning. 

Let's print the dataframe at the end so the user better grasps the expected content and formatting ?

docs/runnable_examples/05_solution_verification/plot_tan_open_hole_convergence.py
```diff
@@ -0,0 +1,430 @@
...
+# WARRANTIES WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED
+# WARRANTIES OF MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL
+# THE AUTHOR BE LIABLE FOR ANY SPECIAL, DIRECT, INDIRECT,
+# OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES WHATSOEVER RESULTING
+# FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN ACTION OF CONTRACT,
+# NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF OR IN CONNECTION
+# WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+
+"""
+Solution verification of the analytic Tan open-hole model
```
FlorentGRT, line 18: @Seb, j'ai fait pas mal de commentaires au début sur la façon d'aborder cette page de la doc, on peut en discuter de vive voix si tu veux. Tout reste est super.

docs/how_to/model_integration.md
FlorentGRT: "Integrating" ?

interest => singular ?

docs/how_to/model_integration.md
```diff
@@ -21,29 +21,47 @@ To learn more about the advantages of this approach, you can refer to
...
+--8<-- "src/vimseo/problems/mock/mock_pre_run_post/mock_components_lc1.py"
+```
+
+If you also want to define a material for this model, you can look at the below example:
 
 ```python
 --8<-- "src/vimseo/problems/mock/mock_pre_run_post/mock_with_material.py"
 ```
+
+The material itself is defined from two files. Its grammar, possibly defining the bounds and types of the properties:
```
FlorentGRT, line 59: It is weird that we talk about material grammar without before mentionning model/components grammar.

docs/how_to/model_integration.md
FlorentGRT, line 14: This page is a nice start, we should go even further, I think I have a big role to play here, but not in this MR. This is something we should tackle early on in SUNRISE/DUCTILE

docs/references/open_hole_tan_model/index.md
```diff
@@ -0,0 +1,28 @@
...
+
+ This work is licensed under the Creative Commons Attribution-ShareAlike 4.0
+ International License. To view a copy of this license, visit
+ http://creativecommons.org/licenses/by-sa/4.0/ or send a letter to Creative
+ Commons, PO Box 1866, Mountain View, CA 94042, USA.
+-->
+
+# Open Hole Plate Model (Tan Model)
+
+Documentation describing the [**analytical Tan model**](@tan2017stress) used to compute stress fields
```
FlorentGRT: broken link

src/vimseo/problems/tan_oh/tan_oht.py
SebastienBocquet: It is not currently possible to specify a SimpleGrammar, because we are using a material defined as a JSONGrammar. Reason explained below.
But I agree that this is a dangerous behaviour. So we should not allow the user to swith of grmmar, or raise a clear error. We can fix that in a dedicated MR.

Setting default_grammar_type = "SimpleGrammar" on TanRun_Tension / PostFieldExtraction breaks two things, one fatal and one silent:

1. Crash at construction — update_from_file doesn't exist on SimpleGrammar.

Both components are built with material_grammar_file=MATERIAL_GRAMMAR_FILE and BaseComponent.__init__ merges it in via:

self.input_grammar.update_from_file(material_grammar_file)

update_from_file is a JSONGrammar-only method (it parses the JSON-schema file at src/vimseo/material_lib/plane_orthotropic_ply_grammar.json). SimpleGrammar has no such method:

SimpleGrammar has update_from_file: False
JSONGrammar has update_from_file: True
So with SimpleGrammar the component would raise AttributeError: 'SimpleGrammar' object has no attribute 'update_from_file' as soon as it's instantiated.

2. Even if that call were bypassed, bounds would silently disappear.

TanOpenHole.__init__ derives lower_bounds / upper_bounds / input_space only when the first component's grammar is a JSONGrammar:

if isinstance(self._chain.disciplines[0].input_grammar, JSONGrammar):
    self.lower_bounds, self.upper_bounds, self.input_space = load_input_bounds(...)
else:
    ...  # everything defaults to [-inf, inf]

docs/references/materials/ta6v.md
SebastienBocquet: The new plane orthotropic ply material is added

docs/references/open_hole_tan_model/validity.md
SebastienBocquet: Your suggestion is added

src/vimseo/problems/tan_oh/tan_oht.py
SebastienBocquet: We can start with a tension and add more load cases later.

src/vimseo/problems/tan_oh/tan_oht.py
SebastienBocquet: This is a bug, thanks. It is now fixed and 5 self-consistency tests added based on physical limit cases.

src/vimseo/problems/tan_oh/tan_oht.py
SebastienBocquet: This is a bug, now fixed.

src/vimseo/tools/bayes/bayes_analysis.py
SebastienBocquet: Cleaned

docs/_draft_design/calibration.rst
```diff
@@ -1,5 +1,5 @@
 ..
-    Copyright 2021 IRT Saint Exupéry, https://www.irt-saintexupery.com
+    Copyright 2021 IRT Saint Exupery, https://www.irt-saintexupery.com
 
     This work is licensed under the Creative Commons Attribution-ShareAlike 4.0
```
SebastienBocquet: toto

docs/references/open_hole_tan_model/equations.md
SebastienBocquet: fixed

src/vimseo/material_lib/plane_orthotropic_ply.json
SebastienBocquet: lower bounds set to 0, and large numbers switched to scientific notation.

src/vimseo/material_lib/plane_orthotropic_ply.json
SebastienBocquet: Is it OK to set lower bound to 0 for all variables?

src/vimseo/material_lib/plane_orthotropic_ply.json
SebastienBocquet: Scientific notation used

docs/runnable_examples/05_solution_verification/plot_tan_open_hole_convergence.py
SebastienBocquet: Would `plot_solution_verification_tan_model.py` be OK?

docs/runnable_examples/05_solution_verification/plot_tan_open_hole_convergence.py
SebastienBocquet: I removed four useless occurences of these two lines in the examples.

src/vimseo/problems/tan_oh/tan_oht.py
FlorentGRT: So the short term solution is either to not define additionnal grammar or define a JSON grammar ? A JSON grammar should be trivial and have minimal cost thanks to an agent.

src/vimseo/material_lib/orthotropic.py
FlorentGRT, line 1: this file seems weird and not used

docs/explanations/model_integration/model_integration.md
FlorentGRT, line 52: mention that these models are not in the core pluggin but in the vimseo-composites ? vimseo-abaqus offers similare representation

docs/runnable_examples/05_solution_verification/plot_tan_open_hole_convergence.py
```diff
@@ -0,0 +1,430 @@
...
+=========================================================
+
+Assess the discretization error of the :class:`TanOpenHole` model with the
+``DiscretizationSolutionVerification`` tool, and compare two field-derived
+quantities: one that converges smoothly and one that exhibits a *sawtooth*
+behaviour for which the Richardson extrapolation cannot be computed.
+
+The Tan model provides an *analytic* membrane stress field for a plate with a
+circular hole (see the :ref:`Tan model reference <open-hole-plate-model-tan-model>`).
```
FlorentGRT: the :ref: field does not work : 

Image

docs/runnable_examples/05_solution_verification/plot_tan_open_hole_convergence.py
FlorentGRT: yes

My remark was more about having of tree of directories in `vimseo/doc` mathcing more clearly the structure and names in the compiled doc

docs/how_to/model_integration.md
SebastienBocquet: Done

SebastienBocquet: Fair enough

docs/how_to/model_integration.md
SebastienBocquet: I added that as a an inline comment

docs/references/open_hole_tan_model/equations.md
SebastienBocquet: Fixed

docs/references/open_hole_tan_model/validity.md
SebastienBocquet: Done

docs/references/open_hole_tan_model/index.md
SebastienBocquet: Fixed

docs/runnable_examples/05_solution_verification/plot_tan_open_hole_convergence.py
SebastienBocquet: We discussed in real life that we can postpone this.

docs/runnable_examples/05_solution_verification/plot_tan_open_hole_convergence.py
SebastienBocquet: Ok, I would just keep the description of sigma_xx_probe and sigma_xx_peak.
I would put parapraph 2 as TanModel docstring.
And put the last paragrpah in the SolutionVerification analysis docstring, rephrased to be more generic.

docs/runnable_examples/05_solution_verification/plot_tan_open_hole_convergence.py
SebastienBocquet: I added your proposal in the example docstring.

docs/runnable_examples/05_solution_verification/plot_tan_open_hole_convergence.py
SebastienBocquet: Done

docs/runnable_examples/05_solution_verification/plot_tan_open_hole_convergence.py
SebastienBocquet: Moved to SolutionVerificationTool docstring, which is better suited location.

docs/runnable_examples/05_solution_verification/plot_tan_open_hole_convergence.py
SebastienBocquet: Yes, removed.

docs/runnable_examples/05_solution_verification/plot_tan_open_hole_convergence.py
SebastienBocquet: Removed since the model now takes the grid size.
We can either add a pre-processing allowing to specify a coarsening factor, or a more integrated approach where a coarsening can be set as input of the SolutionVerificationTool.
Ok for writing that in an issue, for a separate MR?

src/vimseo/material_lib/orthotropic.py
SebastienBocquet: Removed

src/vimseo/problems/tan_oh/tan_oht.py
SebastienBocquet: For the moment, we need to use JSONGrammar, because we are using a material, defined as a JSONGRammar. But we need to better handle this, at least raise a clear error if setting a SimpleGrammar, and document the type of grammar to use depending on the situation (using a material or not).

src/vimseo/problems/tan_oh/tan_oht.py
SebastienBocquet: see previous answer.

src/vimseo/problems/tan_oh/tan_oht.py
SebastienBocquet: Finally, after going back and forth, I prefer being explicit, even if it is more verbose. It facilitates debugging and avoids navigating in the files to understand the interface.

src/vimseo/problems/tan_oh/tan_oht.py
SebastienBocquet: Ok, re-implemented more explicitely.

tests/core/test_model_composition.py
SebastienBocquet: Useless, I remove this reference.

docs/runnable_examples/05_solution_verification/plot_tan_open_hole_convergence.py
SebastienBocquet: The API may be simplified by considering the passed data as dicts. It would naturally handles vectors. Also, the SolutionVerification tool knows which data is an input or output, so no need to specify them. Let's handle that in an issue for another MR.

src/vimseo/problems/tan_oh/tan_oht.py
FlorentGRT: we also need an issue for coarsening_factor => mesh_size

src/vimseo/material_lib/plane_orthotropic_ply.json
```diff
@@ -0,0 +1,228 @@
...
+            "lower_bound": 0.0,
+            "upper_bound": 1000000.0
+          }
+        },
+        {
+          "name": "nu12",
+          "description": "Major Poisson's ratio, in-plane, direction 1-2 (dimensionless)",
+          "value": 0.3,
+          "lower_bound": 0.0,
+          "upper_bound": 1.0,
```
FlorentGRT, line 92: maximum is 0.5, not 1.0

src/vimseo/material_lib/plane_orthotropic_ply.json
```diff
@@ -0,0 +1,228 @@
...
+            "lower_bound": 0.0,
+            "upper_bound": 1000000.0
+          }
+        },
+        {
+          "name": "Xc",
+          "description": "Longitudinal compressive strength, fibre direction (MPa)",
+          "value": 1200.0,
+          "lower_bound": -1000000.0,
+          "upper_bound": 1000000.0,
```
FlorentGRT, line 138: from the given value, the convention seem to be that Xc must be positive here => let be clear on that policy by making lower_bound=0.0

Idem for Yc

src/vimseo/material_lib/plane_orthotropic_ply.json
FlorentGRT: quite a loose tolerance but ok

src/vimseo/material_lib/orthotropic.py
FlorentGRT: I still see it in ad6ee79a2d488713f43788552dccf67e9186f693 (commit "review")

docs/explanations/model_integration/model_integration.md
FlorentGRT: still remains

docs/references/open_hole_tan_model/validity.md
```diff
@@ -0,0 +1,27 @@
...
+
+ This work is licensed under the Creative Commons Attribution-ShareAlike 4.0
+ International License. To view a copy of this license, visit
+ http://creativecommons.org/licenses/by-sa/4.0/ or send a letter to Creative
+ Commons, PO Box 1866, Mountain View, CA 94042, USA.
+-->
+
+# Tan model validity
+
+Tan's model is based on an elastic approximation.
```
FlorentGRT, line 12: "Tan's model is based on an elastic approximation. \n However,  when ..." => "TanOpenHole is a model representing composite Open-Hole sample (notched plate), based on Tan's work on the stress-field analytic solution (pure elasticity hypothesis) and and the "point-stress distance" failure criterion. \n When ..."

docs/references/open_hole_tan_model/validity.md
```diff
@@ -0,0 +1,27 @@
...
+However, when damage occurs (matrix cracking, and possibly later some fibers failure), there are some local stress redistributions and the considered linear elastic law b
+becomes invalid locally. Engineers use the "point-stress" method,
```
FlorentGRT, line 14: "b becomes" => "becomes"

docs/runnable_examples/05_solution_verification/plot_tan_open_hole_convergence.py
```diff
@@ -0,0 +1,420 @@
...
+"""
+Solution verification of the analytic Tan open-hole model
+=========================================================
+
+Assess the discretization error of the :class:`TanOpenHole` model with the
+``DiscretizationSolutionVerification`` tool, which is based on the
+[Richardson extrapolation](../../../reference/vimseo/tools/verification/solution_verification.md#richardson-extrapolation)
+[@richardson1911_finite_differences]
+[@krysl2022_confidence_intervals_richardson], and compare two
+field-derived quantities: one that converges smoothly (``sigma_xx_probe``) and one that exhibits a
```
FlorentGRT, line 26: "one that converges smoothly (``sigma_xx_probe``) and one that exhibits a
*sawtooth* behaviour (``sigma_xx_peak``) for which the Richardson extrapolation cannot
be computed:"

you have a double ":" here => integrate the text above in the two items below

docs/runnable_examples/05_solution_verification/plot_tan_open_hole_convergence.py
```diff
@@ -0,0 +1,420 @@
...
+- ``sigma_xx_peak``: the maximum of ``sigma_xx`` over the grid nodes. Because the
+  node that happens to fall closest to the stress concentration jumps around as
+  the grid changes, this quantity is *non-monotone* -- a sawtooth.
+
+Other relevant metrics may also be integrated similarly in a minimal way.
+Because the solution is analytic, the values stored at the grid nodes are exact.
+What a coarse grid degrades is any quantity read from the discrete field. We look
+at two of them, plus the model's own native scalar outputs as a baseline:
+
+This model is used a use-case case, despite having a working principle different from
```
FlorentGRT, line 41: "This model is **here** used a use-case case **for the SolutionVerification tool**, ..." ?

docs/runnable_examples/05_solution_verification/plot_tan_open_hole_convergence.py
```diff
@@ -0,0 +1,420 @@
...
+*sawtooth* behaviour (``sigma_xx_peak``) for which the Richardson extrapolation cannot
+be computed:
+
+- ``sigma_xx_probe``: ``sigma_xx`` bilinearly interpolated at a fixed point just
+  outside the hole. It converges smoothly as the grid is refined.
+- ``sigma_xx_peak``: the maximum of ``sigma_xx`` over the grid nodes. Because the
+  node that happens to fall closest to the stress concentration jumps around as
+  the grid changes, this quantity is *non-monotone* -- a sawtooth.
+
+Other relevant metrics may also be integrated similarly in a minimal way.
```
FlorentGRT, line 36: "Other relevant metrics may also be integrated similarly in a minimal way." => je ne pense pas le lecteur puisse tirer qqch de cette phrase, il vaut mieux la supprimer

docs/runnable_examples/05_solution_verification/plot_tan_open_hole_convergence.py
```diff
@@ -0,0 +1,420 @@
...
+Because the solution is analytic, the values stored at the grid nodes are exact.
```
FlorentGRT, line 37: "Because the solution is analytic, the values stored at the grid nodes are exact.
What a coarse grid degrades is any quantity read from the discrete field."

C'est rude, le lecteur manque de contexte sur le fonctionnement du modèle, il ne sait même pas que le modèle fournit un champ ni que ceci est fait à partir d'une solution analytique. Je te propose plutôt qqch de ce genre :

"From the analytic solution, the model computes grid nodes evaluations and interpolate a field. Any value of at a grid node is therefore exact, but corse meshing degrades any other value querried from the discrete field".

docs/runnable_examples/05_solution_verification/plot_tan_open_hole_convergence.py
```diff
@@ -0,0 +1,420 @@
...
+        # Avoid duplicating the shared legend entries in the right-hand panel.
+        trace.showlegend = False
+        combined.add_trace(trace, row=1, col=2)
+    combined.update_xaxes(title_text="dx")
+    combined.update_yaxes(title_text=y_title, row=1, col=1)
+    return combined
+
+
+# %%
+# None of the four quantities of interest -- the two field-derived ones and the
```
FlorentGRT, line 108: remplacer "-- ...--" par " ( ...) "

src/vimseo/problems/tan_oh/tan_oht.py
```diff
@@ -0,0 +1,509 @@
...
+    """An Open Hole model based on Tan theory.
+
+    The model provides an *analytic* membrane stress field for a plate with a
+    circular hole (see the :ref:`Tan model reference
+    <open-hole-plate-model-tan-model>`). The field is nonetheless evaluated on
+    a grid whose resolution is driven by the ``grid_size`` input: the grid has
+    ``grid_size`` points per direction, so a larger ``grid_size`` gives a
+    finer grid. Because the solution is analytic, the values stored at the grid nodes
+    are exact. What a coarse grid degrades is any quantity post-processed from the
+    grid values.
```
FlorentGRT, line 336: this definition-usage of "grid_size" is very problematic :

1.  The model input should refer to the spacing between two nodes (like mesh_size), not the number of nodes. 
2.  "grid_size" is very equivocal for this current usage

I propose to do 1. , and keep "grid_size" naming. I am also ok with "grid_spacing".

tests/problems/test_tan_oht.py
```diff
@@ -0,0 +1,417 @@
...
+# (handled by the root-separation floor in ``Calc_S12_eff``). Both are needed for
+# this reference value to be deterministic and reproducible across BLAS/platforms.
+QUASI_ISOTROPIC_STACKING = np.array([0.0, 45.0, -45.0, 90.0, 90.0, -45.0, 45.0, 0.0])
+
+# A strongly orthotropic (0-dominant) stacking. Its roots s1 and s2 are well
+# separated (|s1 - s2| ~ 4.2), so the solution is well conditioned and
+# reproducible across platforms (see the module docstring of ``tan_lib``).
+ORTHOTROPIC_STACKING = np.array([0.0, 0.0, 90.0, 0.0, 0.0, 90.0, 0.0, 0.0])
+
+# A non-isotropic, well-conditioned orthotropic stiffness (C11 != C22, no
```
FlorentGRT, line 41: heavy, please reduce this comment

src/vimseo/material_lib/plane_orthotropic_ply_grammar.json
FlorentGRT: still TODO
`RF = 1/crit`    (ça serait bien d'avoir les deux métriques RF et crit en sortie du modèle)

- si on regarde directement au point `x=R+d0` (uniquement pertinent en traction/compression pure et si un pli à 0° est effectivement présent), voici la formule simplifiée :
    - `crit = sigmaXX/Xt`   si sigmaXX>0 sinon  `crit =sigmaXX)/Xc`
- si on regarde tout le champ (sauf bien sur la partie r<R+d0), pour n'importe quel cas, voici la formule globale :
    - `crit = max ( abs(max(sigmaXX)/Xt), abs(min(sigmaXX)/Xc))`

Il faut aussi supprimer Yt et Yc

src/vimseo/problems/tan_oh/tan_oht.py
FlorentGRT: I let you handle the possible issue

docs/runnable_examples/05_solution_verification/plot_tan_open_hole_convergence.py
SebastienBocquet: Done

docs/runnable_examples/05_solution_verification/plot_tan_open_hole_convergence.py
SebastienBocquet: Removed

docs/runnable_examples/05_solution_verification/plot_tan_open_hole_convergence.py
SebastienBocquet: Done

docs/runnable_examples/05_solution_verification/plot_tan_open_hole_convergence.py
SebastienBocquet: Taken into account, and placed in a note.

docs/runnable_examples/05_solution_verification/plot_tan_open_hole_convergence.py
SebastienBocquet: Done

src/vimseo/material_lib/orthotropic.py
SebastienBocquet: Indeed, removed

src/vimseo/material_lib/plane_orthotropic_ply.json
SebastienBocquet: Fixed,a nd also fixed in the grammar

src/vimseo/material_lib/plane_orthotropic_ply.json
SebastienBocquet: Fixed, and also in the grammar

src/vimseo/problems/tan_oh/tan_oht.py
SebastienBocquet: Done

tests/problems/test_tan_oht.py
SebastienBocquet: Done

tests/problems/test_tan_oht.py
SebastienBocquet: It does use `compute_c_strat`

src/vimseo/material_lib/plane_orthotropic_ply_grammar.json
SebastienBocquet: Done, only for the XX.

src/vimseo/problems/tan_oh/tan_oht.py
SebastienBocquet: I went for simple fix: renaming `grid_size` to `grid_resolution`. We can complexify later with a `dx`, `dy` inputs.

PR #30, opened 2026-03-26
Serialize result of tools to hdf5

SebastienBocquet:
Tool result was writeen on disk and reloaded using pickle.
Now, the methods ``BaseTool.load_results`` and ``BaseTool.save_results()`` use hdf5 by default.
Pickle is still supported.

Note that specifically for the ``SpaceTool``, a serialization to json is implemented. So json is also supported as a serialization format, but will only work for a ``SpaceTool`` result.


src/vimseo/core/model_description.py
```diff
@@ -76,7 +76,7 @@ def _get_multiline(self):
                 dumps(
                     self.default_inputs[group_name],
                     sort_keys=True,
-                    indent=4,
+                    # indent=4,
```
MonPetitCanardArcEnCiel: Please do not leave commented code

src/vimseo/tools/io/base_reader_file.py
```diff
@@ -24,6 +24,7 @@
 class BaseReaderFileSettings(BaseSettings):
     file_name: str | Path = ""
     directory_path: str | Path = ""
+    tool_name: str = ""
 
 
 class StreamlitBaseReaderFileSettings(BaseSettings):
```
MonPetitCanardArcEnCiel: Consider moving these classes to their own files. For instance 
`tools/io/settings/base_file_reader_settings.py` and `tools/io/settings/base_streamlit_file_reader_settings.py`.

MonPetitCanardArcEnCiel: Consider renaming this class as `BaseStreamlitFileReaderSettings`. Is this base class necessary since it does not hav any setting of its own ? Shouldn't it inherit from `BaseReaderFileSettings` ?

src/vimseo/tools/io/base_reader_file.py
```diff
@@ -24,6 +24,7 @@
 class BaseReaderFileSettings(BaseSettings):
```
MonPetitCanardArcEnCiel: Consider renaming this class as `BaseFileReaderSettings`.

src/vimseo/tools/io/reader_file_result.py
```diff
@@ -36,40 +35,24 @@ class ReaderFileToolResult(BaseReaderFile):
...
+
     def __init__(
         self,
         tool_name: str,
     ):
         super().__init__()
         self.__tool_name = tool_name
-        self.__io = IOFactory().create(f"{tool_name}FileIO")
-        self.result = ToolResultsFactory().create(tool_name)
+        self.result = AnalysisToolsFactory().create(tool_name).result
```
MonPetitCanardArcEnCiel, line 47: The attribute being assigned here is `result` (singular) but at line 32, you declare attribute `results` (plural). Is this intentional ?

MonPetitCanardArcEnCiel, line 47: You recreate `ReaderFileToolResult.result` the exact same way in `ReaderFileToolResult.execute`.
Is it necessary to create it here ? Consider initializing it as `None` or directly as an empty object instead.

src/vimseo/tools/io/reader_file_result.py
MonPetitCanardArcEnCiel: The order of the words does not make much sense to me. Consider renaming the class as `ResultFileReader`, `ResultFileReaderTool` or `ResultsFromFileTool`.

src/vimseo/tools/io/base_reader_file.py
```diff
@@ -24,6 +24,7 @@
...
     file_name: str | Path = ""
     directory_path: str | Path = ""
+    tool_name: str = ""
```
MonPetitCanardArcEnCiel: Consider rewriting these attributes as pydantic `Field` objects.

src/vimseo/tools/base_result.py
```diff
@@ -45,10 +62,57 @@ class BaseResult(metaclass=GoogleDocstringInheritanceMeta):
 
     def to_pickle(self, file_path):
         """Save result instance to disk."""
-        file_path = f"{file_path}.pickle"
         with Path(file_path).open("wb") as f:
             pickle.dump(self, f)
 
+    # TODO add an entry point to open myhdf5?
+    def to_hdf5(self, path: str | Path) -> None:
+        """Serialization of a BaseResult to hdf5 file.
```
MonPetitCanardArcEnCiel:
```suggestion
        """Serialize a BaseResult into an HDF5 file.
```

src/vimseo/tools/base_result.py
```diff
@@ -45,10 +62,57 @@ class BaseResult(metaclass=GoogleDocstringInheritanceMeta):
...
+        like <https://myhdf5.hdfgroup.org/>.
+        For more information about hdf5, see
+        [HDF5 documentation](https://docs.hdfgroup.org/hdf5/develop/)."""
+        with h5py.File(path, "w") as f:
+            f.attrs["__class__"] = type(self).__name__
+            for fld in fields(self):
+                serialize_value(f, fld.name, getattr(self, fld.name))
+
+    def to_hdf5_buffer(self) -> BytesIO:
+        """Serialization of a BaseResult to a hdf5 buffer."""
```
MonPetitCanardArcEnCiel:
```suggestion
        """Serialize a BaseResult into an HDF5 buffer."""
```

src/vimseo/tools/base_result.py
```diff
@@ -60,3 +124,46 @@ def save_metadata_to_disk(self, file_path: Path | str = ""):
 
     def __post_init__(self):
         self.metadata = ToolResultMetadata()
+
+
+def assert_results_equal(r1, r2):
+    """Compare récursivement deux dataclasses/dicts/arrays."""
```
MonPetitCanardArcEnCiel:
```suggestion
    """Recursively compare two dataclasses/dicts/arrays."""
```

src/vimseo/tools/base_result.py
```diff
@@ -60,3 +124,46 @@ def save_metadata_to_disk(self, file_path: Path | str = ""):
...
+    if dataclasses.is_dataclass(r1):
+        assert type(r1) is type(r2)
+        for fld in dataclasses.fields(r1):
+            if fld.name == "generic":
+                continue
+            print("Compared field:", fld.name)
```
MonPetitCanardArcEnCiel: Consider using a logger object rather than a `print` method. This should raise a warning on your IDE.

src/vimseo/tools/base_result.py
```diff
@@ -60,3 +124,46 @@ def save_metadata_to_disk(self, file_path: Path | str = ""):
...
+            assert_results_equal(getattr(r1, fld.name), getattr(r2, fld.name))
+    elif isinstance(r1, matplotlib.figure.Figure):
+        # Figures are not comparable, only the type is checked
+        assert type(r1) is type(r2)
```
MonPetitCanardArcEnCiel: Consider raising a warning or at least writing a log message saying this.

src/vimseo/tools/base_result.py
```diff
@@ -60,3 +124,46 @@ def save_metadata_to_disk(self, file_path: Path | str = ""):
...
+    elif isinstance(r1, pd.DataFrame):
+        pd.testing.assert_frame_equal(r1, r2)
+    elif isinstance(r1, dict):
+        assert set(r1.keys()) == set(r2.keys())
+        for k in r1:
+            assert_results_equal(r1[k], r2[k])
+    elif isinstance(r1, list):
+        assert len(r1) == len(r2)
+        for v1, v2 in zip(r1, r2, strict=False):
+            assert_results_equal(v1, v2)
```
MonPetitCanardArcEnCiel, line 155: Not type testing fro `r2` in these cases ?

src/vimseo/tools/base_tool.py
MonPetitCanardArcEnCiel, line 83: Consider moving these to their own modules

MonPetitCanardArcEnCiel, line 83: Consider rewriting the attributes as pydantic `Field` objects.

MonPetitCanardArcEnCiel: Except if they are only used in this file and are not meant for the en user to use ? In this case may be prefix these with an `_` ? We could ask Antoine what he thinks about this.

src/vimseo/tools/base_tool.py
```diff
@@ -406,24 +434,34 @@ def load_options_from_metadata(self, file_path: str | Path | None = None):
...
 
-    # TODO add file_format option, to call to_csv method in the results.
-    # TODO add save_intermediate option to call a save_intermediate method
-    #  which saves the intermediate nodes, or activate the save_results method
-    #  in the computation functions?
-    def save_results(self, prefix: str = "") -> None:
+    def save_results(self, prefix: str = "", file_format="hdf5") -> None:
         """Save the results of the tool on disk. The file path is
-        :attr:`working_directory` / ``{filename}_result.pickle``.
+        :attr:`working_directory` / ``{filename}_result.{file_format}``.
```
MonPetitCanardArcEnCiel:
```suggestion
       `BaseTool.working_directory` / ``{filename}_result.{file_format}``.
```

I don't think this is the correct usage of `:attr:` (I don't know what the correct usage is). This does not render in the doc

MonPetitCanardArcEnCiel: There are others (and `:meth:` as well) but we could fix this in a future PR.

src/vimseo/tools/serializer.py
```diff
@@ -0,0 +1,312 @@
...
+
+from vimseo.utilities.json_grammar_utils import EnhancedJSONEncoder
+
+if TYPE_CHECKING:
+    import h5py
+
+LOGGER = logging.getLogger(__name__)
+
+
+def to_dataframe(value, key, group, type_name):
```
MonPetitCanardArcEnCiel: Please typehint the arguments (except if the method is meant to be private (it might still be a good idea to typehint though))

MonPetitCanardArcEnCiel: This method does not seem to be used anywhere else. If it is not meant to be used outside of this module, consider prefixing with `__` to indicate it is private.

src/vimseo/tools/serializer.py
```diff
@@ -0,0 +1,312 @@
...
+            cols_group.attrs[f"__type__{col_key}"] = "json_col"
+        else:
+            cols_group.create_dataset(col_key, data=col_data.values)
+            # Add human-readable column name as attribute
+            cols_group[col_key].attrs["column_name"] = (
+                json.dumps(list(col)) if is_multiindex else str(col)
+            )
+
+
+def from_dataframe(item):
```
MonPetitCanardArcEnCiel: Same remark about privacity

MonPetitCanardArcEnCiel: Is `item` supposed to be `DataFrame` or sub`DataFrame`, or another `pandas` object, or someting else ? Consider using a more specific name.

src/vimseo/tools/serializer.py
```diff
@@ -0,0 +1,312 @@
...
+        sub = group.require_group(key)
+        sub.attrs["__type__"] = "list"
+        sub.attrs["__len__"] = len(value)
+        for i, v in enumerate(value):
+            serialize_value(sub, str(i), v)
+
+    # TODO handle distributions and parameter_space in a second time
+    # elif isinstance(value, DistributionImplementation):
+    #     group.attrs[key] = value.settings.model_dump_json()
+    #     group.attrs[f"__type__{key}"] = "distribution"
```
MonPetitCanardArcEnCiel: Please remove commented code

src/vimseo/tools/serializer.py
```diff
@@ -0,0 +1,312 @@
...
+    elif isinstance(value, dict):
+        if key in group:
+            del group[key]
+        sub = group.create_group(key)
+        sub.attrs["__type__"] = "dict"
+        sub.attrs["__empty__"] = len(value) == 0
+        for k, v in value.items():
+            try:
+                serialize_value(sub, str(k), v)
+            except Exception:  # noqa: BLE001
```
MonPetitCanardArcEnCiel: Isn't this already taken care of in the `else` block on line 178 ?

src/vimseo/tools/serializer.py
```diff
@@ -0,0 +1,312 @@
...
+        df = from_dataframe(item)
+        return Dataset.from_dataframe(df)
+
+    if type_ == "dataframe":
+        return from_dataframe(item)
+
+    # TODO handle distributions and parameter_space in a second time
+    if type_ == "distribution":
+        settings = DistributionSettings(**item)
+        return type(settings.distribution_name)(settings=settings)
```
MonPetitCanardArcEnCiel: The corresponding part in `serialize_value` was commented out. Was this supposed to be removed ? Currently, there is no way to serialize a distribution, if I am not mistaken.

src/vimseo/tools/serializer.py
```diff
@@ -0,0 +1,312 @@
...
+
+    if type_ == "dict":
+        # Empty dict
+        if item.attrs.get("__empty__", False):
```
MonPetitCanardArcEnCiel: Is it possible that a serialized `dict` have no `__empty__` attributes ? Would it be relevant to raise an error if it is not present ?

src/vimseo/tools/serializer.py
```diff
@@ -0,0 +1,312 @@
...
+                elif fld.default_factory is not dataclasses.MISSING:
+                    kwargs[fld.name] = fld.default_factory()
+                else:
+                    kwargs[fld.name] = None
+
+        # Avoid __post_init__ for all dataclasses, otherwise the settings field is set
+        # to default value.
+        obj = cls.__new__(cls)
+        for k, v in kwargs.items():
+            object.__setattr__(obj, k, v)
```
MonPetitCanardArcEnCiel, line 287:
```suggestion
            obj.__setattr__(obj, k, v)
```

src/vimseo/utilities/datasets.py
```diff
@@ -182,6 +181,8 @@ def plot_dataset_element(
      Returns:
          The plot result.
     """
+    from vimseo.tools.post_tools.line_plot import Lines
```
MonPetitCanardArcEnCiel, line 184: Why import this here rather than at the module level ?

src/vimseo/tools/serializer.py
```diff
@@ -0,0 +1,312 @@
...
+            class_parts = parts[i:]
+            try:
+                module = importlib.import_module(module_name)
+                obj = module
+                for part in class_parts:
+                    obj = getattr(obj, part)
+                cls = obj
+                break
+            except (ModuleNotFoundError, AttributeError):
+                continue
```
MonPetitCanardArcEnCiel, line 265: Why not use an attribute `__module__` to directly store the module name ?

For retrocompatibility, you could keep this block only when no attribute `__module__` is present in `item.attrs`.

src/vimseo/tools/serializer.py
```diff
@@ -0,0 +1,312 @@
...
+            value = bool(value)
+        elif isinstance(value, (np.str_, str)):
+            value = str(value)
+        group.attrs[key] = value
+        group.attrs[f"__type__{key}"] = "primitive"
+
+    elif dataclasses.is_dataclass(value) and not isinstance(value, type):
+        sub = group.require_group(key)
+        sub.attrs["__type__"] = "dataclass"
+        sub.attrs["__class__"] = f"{type(value).__module__}.{type(value).__qualname__}"
```
MonPetitCanardArcEnCiel, line 162: Why not use an attribute `__module__` to help when deserializing ?
```suggestion
        sub.attrs["__class__"] = f"{type(value).__module__}.{type(value).__qualname__}"
        sub.attrs["__module__"] = f"{type(value).__module__}"
```

src/vimseo/utilities/generate_validation_reference.py
```diff
@@ -12,6 +12,42 @@
...
 # along with this program; if not, write to the Free Software Foundation,
 # Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+"""
+Generate reference data from a model.
+
+layup = [
+    "90_-30_90_-30_-30_90_-30_90"
+]
+
+outputs_to_bias = {}  # {"max_strength": Bias(mult_factor=1.05)}
```
MonPetitCanardArcEnCiel: please remove commented code

src/vimseo/utilities/generate_validation_reference.py
```diff
@@ -12,6 +12,42 @@
...
+df = generate_reference_from_dataset(
+    model,
+    reference_data,
+    df=df,
+    specific_inputs=specific_inputs,
+    outputs_to_bias=outputs_to_bias,
+)
+df.to_csv("reference_validation_opfmcube_pst.csv", sep=SEP)
+
+"""
```
MonPetitCanardArcEnCiel, line 17: Wouldn't this be better in a dedicated example in the doc rather than in the API docstring ?
```suggestion
"""A module to generate reference data from a model."""
```

src/vimseo/utilities/pytest_conftest.py
MonPetitCanardArcEnCiel, line 1: Wouldn't this file be better in the `tests` folder ? For instance `tests/conftest.py`

MonPetitCanardArcEnCiel (changes requested):
A first small review. I have yet to review the tests and the doc.

src/vimseo/tools/base_result.py
```diff
@@ -60,3 +124,46 @@ def save_metadata_to_disk(self, file_path: Path | str = ""):
 
     def __post_init__(self):
         self.metadata = ToolResultMetadata()
+
+
+def assert_results_equal(r1, r2):
```
MonPetitCanardArcEnCiel, line 129: If I am not mistaken, there is no test checking that this method behave properly. This is a problem especially since it is itself used to validate other tests.

tests/tools/io/test_tool_io.py
```diff
@@ -32,15 +32,12 @@ def test_tool_result_reader():
     result = reader.execute(
         settings=BaseReaderFileSettings(
             file_name=IO_DATA_DIR / "space_tool" / "space_tool_result.json",
+            tool_name=tool_name,
         )
     )
     assert isinstance(result, SpaceToolResult)
```
MonPetitCanardArcEnCiel, line 38: Consider checking that the `result` object contains what it is supposed to contain.

tests/tools/test_serialization.py
```diff
@@ -0,0 +1,423 @@
...
+#         dist: OTDistribution | None = None
+
+#     # Sans serializer custom, doit tomber en pickle
+#     result = ResultWithDist(dist=dist)
+#     rt = roundtrip(result, tmp_hdf5)
+
+#     import h5py
+
+#     with h5py.File(tmp_hdf5, "r") as f:
+#         assert f["dist"].attrs["__type__"] == "distribution"
```
MonPetitCanardArcEnCiel: Please don't leave commented code

tests/tools/test_validation_case.py
```diff
@@ -232,7 +143,113 @@ def test_to_dataframe(tmp_wd):
     point_2.metadata.report["measured_output_names"] = ["y1"]
     result = ValidationCaseResult()
     result.set_from_point_results([point_1, point_2])
+    return result
+
 
+@pytest.fixture
+def deterministic_validation_case(reference_data):
```
MonPetitCanardArcEnCiel, line 150: Please add a docstring to the fixture

tests/utilities/test_datasets.py
```diff
@@ -98,33 +99,30 @@ def test_dataframe_to_dataset_with_duplicated_names(mock_dataset_with_duplicated
         .get_view(variable_names=["a"], group_names=IODataset.INPUT_GROUP)
         .to_numpy()
         .ravel(),
-        df[f"a[{IODataset.INPUT_GROUP}]"].to_numpy(),
+        df["a{" + IODataset.INPUT_GROUP + "}"].to_numpy(),
```
MonPetitCanardArcEnCiel: In an `f-string` you can escape the brackets by using two, like this : `{{`

```suggestion
        df[f"a{{{IODataset.INPUT_GROUP}}}"].to_numpy(),
```

tests/utilities/test_datasets.py
```diff
@@ -98,33 +99,30 @@ def test_dataframe_to_dataset_with_duplicated_names(mock_dataset_with_duplicated
...
     )
     assert_array_equal(
         ds
         .get_view(variable_names=["a"], group_names=IODataset.OUTPUT_GROUP)
         .to_numpy()
         .ravel(),
-        df[f"a[{IODataset.OUTPUT_GROUP}]"].to_numpy(),
+        df["a{" + IODataset.OUTPUT_GROUP + "}"].to_numpy(),
```
MonPetitCanardArcEnCiel:
```suggestion
        df[f"a{{{IODataset.OUTPUT_GROUP}}}"].to_numpy(),
```

tests/utilities/test_datasets.py
```diff
@@ -98,33 +99,30 @@ def test_dataframe_to_dataset_with_duplicated_names(mock_dataset_with_duplicated
...
     )
     assert_array_equal(
         ds
         .get_view(variable_names=["b"], group_names=IODataset.OUTPUT_GROUP)
         .to_numpy()
         .T[0],
-        df[f"b[{IODataset.OUTPUT_GROUP}][0]"].to_numpy(),
+        df["b{" + IODataset.OUTPUT_GROUP + "}[0]"].to_numpy(),
```
MonPetitCanardArcEnCiel:
```suggestion
        df[f"b{{{IODataset.OUTPUT_GROUP}}}[0]"].to_numpy(),
```

MonPetitCanardArcEnCiel (changes requested):
This is the review for the tests. I still the examples to review.

tests/tools/test_serialization.py
```diff
@@ -0,0 +1,423 @@
...
+
+class TestNumpyArrays:
+    def test_1d_array(self, tmp_hdf5):
+        arr = np.array([1.0, 2.0, 3.0])
+        result = BayesAnalysisResult(raw_samples=arr)
+        rt = roundtrip(result, tmp_hdf5)
+        np.testing.assert_array_equal(rt.raw_samples, arr)
+
+    def test_2d_array(self, tmp_hdf5):
+        arr = np.random.rand(10, 4)  # noqa: NPY002
```
MonPetitCanardArcEnCiel: This is deprecated. Is the randomness necessary ? If so consider using `numpy.random.default_rng()` to generate a `BitGenerator` object and then calling `BitGenerator.random(size=(10,4))`

Else : 

```suggestion
        arr = np.linspace(0,4, 40).reshape((10,4))
```

src/vimseo/core/model_description.py
SebastienBocquet: Fixed

src/vimseo/tools/io/base_reader_file.py
SebastienBocquet: Done. I have put both settings in the same module, because the way to manage these settings for Streamlit should be changed.
I will open an issue, because the settings specifically defined for the Streamlit dashboard (for workflow set-up) should not be present at "tool" level. Indeed, a tool should remain agnostic of a workflow and the workflow UI (the workflow engine is currently gemseo but others could be added in the future). 
A tool is wrapped into a workflow node (class `WorkflowStep`) to obtain a workflow, and I think the Streamlit setting should be defined there.

SebastienBocquet: Regarding the renaming, the name of settings for the Streamlit workflow dashboard are the name of the corresponding settings prefixed by `Streamlit`. So I propose to leave as is.

Regarding the fact that `StreamlitBaseReaderFileSettings` has no field, it is intentional. In the workflow dahsboard, the file name, tool name, directory path are not necessary because the tool result file is uploaded by the user.

src/vimseo/tools/io/base_reader_file.py
SebastienBocquet: I agree, it is more clear. Are you OK for renaming `BaseReaderFile` into `BaseFileReader`?

src/vimseo/tools/io/reader_file_result.py
SebastienBocquet: No, it is a typo. I fix it.

src/vimseo/tools/serializer.py
SebastienBocquet: Done

src/vimseo/tools/serializer.py
SebastienBocquet: Thanks, I now use:

```
        sub.attrs["__module__"] = f"{type(value).__module__}"
        sub.attrs["__class__"] = f"{type(value).__name__}"
```

src/vimseo/utilities/generate_validation_reference.py
SebastienBocquet: Done

src/vimseo/utilities/generate_validation_reference.py
SebastienBocquet: Yes, I will write an example.

src/vimseo/utilities/pytest_conftest.py
SebastienBocquet: in GEMSEO, a pytest_conftest.py file is in the utils. I guess it is the best practice.

src/vimseo/tools/io/base_reader_file.py
MonPetitCanardArcEnCiel: I'm not sure I understood but I'll trust you. You'll explain to me face to face :)

SebastienBocquet:
I will open an issue to handle retro-compatiblity of tool result loading

Also, the coverage has decreased, so I need to look at it.

src/vimseo/tools/io/base_reader_file.py
MonPetitCanardArcEnCiel: There are a lot of objects that have `ReaderFile` in their names that should be changed to `FileReader`, and same for their modules, but I think this is out of scope for this PR. Let us open a separate issue.

src/vimseo/tools/base_result.py
SebastienBocquet: You are right. One possibility I see is to generate mock results, and alter some fields and check that the difference is detetced. Are you Ok with this approach?

src/vimseo/tools/base_tool.py
MonPetitCanardArcEnCiel: From what I gather this is a remnant of the sphinx documentation and is not used with mkdocs

src/vimseo/tools/base_result.py
MonPetitCanardArcEnCiel: Yes. Or simply check each field that it contains what we expect it contains, whichever is easier

tests/tools/test_validation_case.py
SebastienBocquet: Done

tests/tools/test_serialization.py
SebastienBocquet: Removed

src/vimseo/tools/io/reader_file_result.py
SebastienBocquet: Done

src/vimseo/tools/io/reader_file_result.py
SebastienBocquet: I think it is not exactly the same. In `execute`, I create the corresponding tool in order to load the result.
In the constructor, the result is intialized as a result corresponding to the tool `tool_name`, with its fields being left to default values. I think that I use the fact that this attribute is a result, not `None`, in the workflow dashboard.
I propose to open an issue to assess if the result attr can be set to `None` in a separate MR. If OK with you, I will open an issue.

src/vimseo/tools/io/base_reader_file.py
SebastienBocquet: Done

src/vimseo/tools/base_result.py
SebastienBocquet: Done

src/vimseo/tools/base_result.py
SebastienBocquet: Ok, done

src/vimseo/tools/base_result.py
SebastienBocquet: I think we have to choose which of `r1` or `r2` to use to select the type of comparison to perform, according to its type. Then, the type of the other result is implicitly checked by the logic of the comparison impemented in each case. 
Do you see a case where we could miss a difference between the two results?

src/vimseo/tools/base_tool.py
SebastienBocquet: I propose to address the tool settings for the workflow dashboard in a separate MR. Indeed, these settings should be move to the workflow sub package.

SebastienBocquet: Done (only for `ToolConstructorSettings`)

src/vimseo/tools/base_tool.py
SebastienBocquet: I agree

src/vimseo/tools/serializer.py
SebastienBocquet: Done

SebastienBocquet: Done

src/vimseo/tools/serializer.py
SebastienBocquet: Done

SebastienBocquet: `item` is a `h5py.Group`. With the typehint, I think it helps to understand what it repsresents:
`def __from_dataframe(item: h5py.Group) -> pd.DataFrame:`
Are you OK with this?

src/vimseo/tools/serializer.py
SebastienBocquet: Done

src/vimseo/tools/serializer.py
SebastienBocquet: Done

src/vimseo/tools/serializer.py
SebastienBocquet: I remove this try/except because it does not pass in the except in any test.

src/vimseo/tools/serializer.py
SebastienBocquet: Serialization of a dict defines an attribute `__empty__`. I think the logic is OK. Do you agree?

src/vimseo/tools/serializer.py
SebastienBocquet: I think for this very particular usage (avoiding to call the `__post_init` method), `object.__setattr__(obj, k, v)` is the appropriate syntax

src/vimseo/tools/base_result.py
MonPetitCanardArcEnCiel: The problem I see is rather that if the two types are not compatible, `assert_results_equal` may raise an exception that is not an `AssertionError`. Maybe it's not a big deal, and maybe it is the desired behavior, but I think it may make debugging more difficult.

(for instance if `r1` is a `dict` and `r2 ` a `list` or an `array`)

src/vimseo/utilities/datasets.py
SebastienBocquet: Otherwise I get a circular import. I would need to split the `datasets` module. Can we handle that in another MR?

SebastienBocquet:
I took all your remarks into account, except ading tests on `assert_results_equal`. I will take care of this soon.

src/vimseo/tools/io/base_reader_file_settings.py
```diff
@@ -12,3 +12,29 @@
...
+
+    file_name: str | Path = Field(
+        default="", description="The name of the file to read."
+    )
+    directory_path: str | Path = Field(
+        default="", description="The path to the directory containing the file."
+    )
+    tool_name: str = Field(
+        default="", description="The name of the tool to use for reading the file."
+    )
```
MonPetitCanardArcEnCiel, line 36: Are you sure you want an empty string by default ? Is this handled later ? Isn't it better to make the arguments mandatory ?

src/vimseo/tools/base_result.py
```diff
@@ -45,10 +62,57 @@ class BaseResult(metaclass=GoogleDocstringInheritanceMeta):
 
     def to_pickle(self, file_path):
```
MonPetitCanardArcEnCiel:
```suggestion
    def to_pickle(self, file_path: str | Path):
```

src/vimseo/tools/serializer.py
MonPetitCanardArcEnCiel: Then why not :

```suggestion
        if item.attrs["__empty__"]:
```

It is not very important. I think the logic is OK but the way implies that it is possible that there be no `__empty__` attribute

src/vimseo/tools/serializer.py
MonPetitCanardArcEnCiel: Oh yes I misread. I thought it was a typo. My bad.

docs/runnable_examples/07_validation_case/plot_generate_synthetic_data.py
```diff
@@ -0,0 +1,121 @@
...
+# WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+
+"""# Generating synthetic model output data for examples, tests or mock studies.
+
+## Problem
+
+As a developer, you want ot generate reference data for examples or tests, controlling
+the bias of the data with respect to the model output.
+As a user, you want to generate synthetic experimental data that could be used
+to validate or verify a model. from a model.
```
MonPetitCanardArcEnCiel:
```suggestion
to validate or verify a model.
```

src/vimseo/tools/serializer.py
MonPetitCanardArcEnCiel: Yes !

src/vimseo/tools/io/base_reader_file_settings.py
SebastienBocquet: This may be challenged of course, but currently, the result of a tool is instanciated in the constructor of the tool, with default values. The result field are then set in the `execute` of the tool. So we need default values for each field of a result. It explains these empty strings.

src/vimseo/tools/base_result.py
MonPetitCanardArcEnCiel: I created a distinct PR

src/vimseo/tools/base_result.py
MonPetitCanardArcEnCiel: This will be adressed in the dedicated PR

tests/tools/io/test_tool_io.py
MonPetitCanardArcEnCiel: I opened a dedicated PR.

MonPetitCanardArcEnCiel (approved):

MonPetitCanardArcEnCiel (approved):
All test passed

PR #46, opened 2026-04-20
clean up useless logic in external software subprocess, and ensure the external software is killed in any case.

SebastienBocquet:
When an exception occurs during solver run, there is no mechanism ensuring that the solver subprocesses are killed

Also, useless logic exists in the execution of the solver subprocess:

- explicit logic for convergence log fetching is useless, since naturally hanled by the base_executor.execute logic of log fetching 
- explicit detection whether the subprocess is finished is useless, since it is already detected by subprocess.poll()

### DIFF
--- a/src/vimseo/job_executor/base_executor.py
+++ b/src/vimseo/job_executor/base_executor.py
@@ -1,3 +1,18 @@
+# Copyright 2021 IRT Saint Exupery, https://www.irt-saintexupery.com
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Lesser General Public
+# License version 3 as published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this program; if not, write to the Free Software Foundation,
+# Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+
 # Copyright 2021 IRT Saint Exupéry, https://www.irt-saintexupery.com
 #
 # This program is free software; you can redistribute it and/or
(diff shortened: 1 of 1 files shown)


src/vimseo/job_executor/base_executor.py
```diff
@@ -134,6 +149,9 @@ def options(self) -> dict:
     def command_line(self):
         return self._command_line
 
+    def _terminate_external_software(self):
```
MonPetitCanardArcEnCiel, line 163: shouldn't this be defined as an abstract method ?

SebastienBocquet: Good question. I am not sure that I can implement a generic way of terminating the process. I chatted with Claude Sonnet 4.6, et even after some iterations, I did not manage to get a functional generic termination logic, that I can understand and maintain later. But I am not saying that it is not feasible. But still, I do not know how to test that.

In any case, as-is it is not correct, because there is a log saying that the solver will terminate, but the termination method can be empty. A quick fix could be to rephrase the log, and add a log in the base termination method saying that no termination is implemented, and that some subprocesses can still be running.

MonPetitCanardArcEnCiel: Could the signature not be something like 

```
@abstract_method
def _terminate_external_software(self, *args, **kwargs):
```
?
I don't know if this would work, or if this would be best practice. We could ask Antoine when he is back.

As-is, there is just a method that does nothing than can be called.
At least, consider raising `NotImplementedError` in the method in the base class

src/vimseo/job_executor/base_executor.py
MonPetitCanardArcEnCiel, line 1: Consider removing `base` from the file name as I feel it suggest that the class it defines should be an abstract class.
Example : `job_executor.py`

MonPetitCanardArcEnCiel, line 73: Consider removing `Base` from the class name, as I feel it suggests that the class should be abstract, which it isn't.

```suggestion
class JobExecutor(metaclass=GoogleDocstringInheritanceMeta):
```

MonPetitCanardArcEnCiel (changes requested):

PR #47, opened 2026-04-20
refactor ise curve metric

SebastienBocquet:
The implementation of the metric comparing curves was difficult to read and maintain.
Also, the logic was not robust to non-monotonix curve x-axis.

The tests on this metric were not intuitive. Hopefully, it is improved and at least better documented.

### DIFF
--- a/src/vimseo/tools/calibration/calibration_metrics.py
+++ b/src/vimseo/tools/calibration/calibration_metrics.py
@@ -16,7 +16,7 @@
 from __future__ import annotations
 
 import logging
-import uuid
+from datetime import datetime
 from typing import TYPE_CHECKING
 
 import matplotlib.patches as patches
@@ -26,11 +26,9 @@ from gemseo_calibration.measures.integrated_measure import IntegratedMeasure
 from gemseo_calibration.measures.mean_measure import MeanMeasure
 from gemseo_calibration.measures.mse import MSE
 from numpy import abs as np_abs
-from numpy import all as np_all
-from numpy import argwhere
+from numpy import argsort
 from numpy import array
 from numpy import atleast_1d
-from numpy import diff
 from numpy import linspace
 from numpy import max as np_max
 from numpy import mean
@@ -48,6 +46,22 @@ if TYPE_CHECKING:
 LOGGER = logging.getLogger(__name__)
 
 EPSILON = 1e-12
+
+
+def _sort_and_align(x, y):
+    """Sort x (and y accordingly), ensuring x is increasing."""
+    idx = argsort(x)
+    return x[idx], y[idx]
+
+
+def _restrict_to_range(x, y, x_min, x_max):
+    """Restrict (x, y) to the range [x_min, x_max]."""
+    mask = (x >= x_min) & (x <= x_max)
+    return x[mask], y[mask]
+
+
+def _scale_xy(x, y, x0, y0, factor_x, factor_y):
+    return (x - x0) * factor_x, (y - y0) * factor_y
 
 
 class RelativeMSE(MeanMeasure):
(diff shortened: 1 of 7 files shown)


src/vimseo/tools/calibration/calibration_metrics.py
```diff
@@ -50,6 +48,22 @@
...
+
+
+def _restrict_to_range(x, y, x_min, x_max):
+    """Restrict (x, y) to the range [x_min, x_max]."""
+    mask = (x >= x_min) & (x <= x_max)
+    return x[mask], y[mask]
+
+
+def _scale_xy(x, y, x0, y0, factor_x, factor_y):
+    return (x - x0) * factor_x, (y - y0) * factor_y
```
MonPetitCanardArcEnCiel, line 64: This will translate by a vector (`x_0`, `y_0`) on top of the scaling. Is this the desired behavior ? 

Else : 
```suggestion
    return (x - x0) * factor_x + x_0, (y - y0) * factor_y + y_0
```

src/vimseo/tools/calibration/calibration_metrics.py
```diff
@@ -96,40 +110,32 @@ def _evaluate_measure(self, model_dataset: DataType) -> float:  # noqa: D102
...
                 factor_x = 1.0 / max(x_ref[-1] - x_ref[0], x_i[-1] - x_i[0])
                 delta_y = max(np_max(y_ref) - np_min(y_ref), np_max(y_i) - np_min(y_i))
                 factor_y = 1.0 / EPSILON if delta_y < EPSILON else 1.0 / delta_y
-                x_ref = scale(x_ref, x0, factor_x)
-                y_ref = scale(y_ref, y0, factor_y)
-                x_i = scale(x_i, x0, factor_x)
-                y_i = scale(y_i, y0, factor_y)
+                x_ref, y_ref = _scale_xy(x_ref, y_ref, x0, y0, factor_x, factor_y)
+                x_i, y_i = _scale_xy(x_i, y_i, x0, y0, factor_x, factor_y)
+
```
MonPetitCanardArcEnCiel, line 135: I'm not sure I like this change : I don't see why grouping the logic of scaling `x` and `y` together, other than removing two lines.

src/vimseo/tools/calibration/calibration_metrics.py
```diff
@@ -138,51 +144,29 @@ def scale(x, x0, factor):
...
-            x_ref_0 = union1d(x_ref, x_i_in_x_ref)
+            for name, value in [("x_left", delta_x_left), ("x_right", delta_x_right)]:
+                mse = MSE(output_name=name)
+                mse.set_reference_data({name: atleast_1d(0.0)})
+                metric = mse._evaluate_measure({name: atleast_1d(value)})
+                (
+                    exceeding_start_metrics
+                    if name == "x_left"
+                    else exceeding_end_metrics
+                ).append(metric)
```
MonPetitCanardArcEnCiel:
```suggestion
                if name == "x_left":
                                exceeding_start_metrics.append(metric)
                else:
                                exceeding_end_metrics.append(metric)
```

I think this is clearer and take less lines

tests/metrics/test_ise_calibration_metric.py
```diff
@@ -26,20 +28,21 @@
...
 from vimseo.utilities.curves_generator import n_points
 
+if TYPE_CHECKING:
+    from numpy import ndarray
+
+
+def _swap_points(x: ndarray, y: ndarray):
+    """Swap two random points in the curve defined by x and y."""
+    idx1 = 0
+    idx2 = 1
```
MonPetitCanardArcEnCiel: This is not random

tests/tools/calibration/test_curve_abscissa_penalization.py
```diff
@@ -34,10 +36,28 @@
...
+# the abscissa range.
+
 
+def create_reference_dataset(
+    x_left: float, x_right: float, y_max: float = 1.0
+) -> IODataset:
+    """Creates a reference dataset for the test.
 
-def create_reference_dataset(x_left: float, x_right: float, y_max: float):
+    The reference curve is generated using the expressions defined in the curves
```
MonPetitCanardArcEnCiel:
```suggestion
    The reference curve is generated using the expressions defined in the curve
```

MonPetitCanardArcEnCiel (changes requested):

tests/metrics/test_ise_calibration_metric.py
MonPetitCanardArcEnCiel: Please remove the `print`

MonPetitCanardArcEnCiel, line 128: Please add an assertion to verify the metric is correctly evaluated.

Consider defining an analytical case, such as two constant curves.

MonPetitCanardArcEnCiel (changes requested):

tests/metrics/test_ise_calibration_metric.py
```diff
@@ -26,20 +28,21 @@
 from vimseo.utilities.curves_generator import get_history
 from vimseo.utilities.curves_generator import n_points
 
+if TYPE_CHECKING:
+    from numpy import ndarray
+
+
+def _swap_points(x: ndarray, y: ndarray):
+    """Swap two random points in the curve defined by x and y."""
```
MonPetitCanardArcEnCiel:
```suggestion
    """Swap two points in the curve defined by x and y."""
```

src/vimseo/tools/calibration/calibration_metrics.py
SebastienBocquet: Yes it is the desired behaviour

src/vimseo/tools/calibration/calibration_metrics.py
SebastienBocquet: I understand your remark. I did that because it was not obvious that the same scaling would be done on x and y. Ideally, we would manipulate a `Curve` object and scale the curve. So manipulating the two arrays at the same time is a low cost 'curve' approach. In the future it would be nice to introduce this curve concept, so I am not sure I want to split the scaling for x and y

tests/metrics/test_ise_calibration_metric.py
SebastienBocquet: Done

SebastienBocquet: Done

MonPetitCanardArcEnCiel (approved):

PR #56, opened 2026-06-26
Coverage

SebastienBocquet:


tests/core/test_model_result.py
```diff
@@ -86,6 +86,28 @@ def test_scalars_to_dataframe():
...
+    result = ModelResult(
+        scalars={"x": 1.0},
+        vectors={"v": array([1.0, 2.0])},
+        curves=[Curve({"xc": array([0.0, 1.0]), "yc": array([0.0, 1.0])})],
+    )
+    text = str(result)
+    assert "Scalars:" in text
+    assert "Vectors:" in text
+    assert "Curves:" in text
+    assert "Fields:" in text
```
MonPetitCanardArcEnCiel: No attribute `field` is filled in the `ModelResult`, yet the test checks for `"Fields"` in the text. Is this the expected behavior ? Are the words "Fields", "Curves", "Vectors" and "Scalars" present in the string representation whether or not the attribute of the `ModelResult` are filled ? If so why fill the attribute in the first place rather than testing the most general behavior ?

MonPetitCanardArcEnCiel: Consider checking that the string representation of the data is present and has the expected format as well

tests/core/test_model_result.py
```diff
@@ -86,6 +86,28 @@ def test_scalars_to_dataframe():
...
+
+
+def test_get_curve_not_found_returns_empty():
+    """Requesting an unknown curve returns an empty list."""
+    result = ModelResult(
+        curves=[Curve({"xc": array([0.0, 1.0]), "yc": array([0.0, 1.0])})],
+    )
+    assert result.get_curve(("unknown_x", "unknown_y")) == []
```
MonPetitCanardArcEnCiel: Consider checking that `result.get_curve` returns the correct curves when requesting a known curve as well.

MonPetitCanardArcEnCiel (changes requested):
A first par of the review

tests/material/test_material.py
```diff
@@ -179,3 +182,91 @@ def test_update_stochastic_material():
...
+    material = Material.from_json(MATERIAL_LIB_DIR / "Ta6v.json")
+    with pytest.raises(ValueError, match="No material property"):
+        material.get_property("Ta6v_elastic_iso", "unknown_property")
+
+
+def test_to_json_schema(tmp_wd):
+    """A material can be exported as a JSON schema file."""
+    material = Material.from_json(MATERIAL_LIB_DIR / "Ta6v.json")
+    material.to_json_schema()
+    assert Path(f"{material.name}_grammar.json").is_file()
```
MonPetitCanardArcEnCiel: Won't this test live spurious files at the root of the repo ? Please check that no generated file remains at the end of the tests (is this what the `tmp_wd` is for ? I can't find it in the pytest doc).

MonPetitCanardArcEnCiel: Consider checking that the content of the json file is compliant.
(except if this is checked elsewhere ?)

tests/material/test_material.py
```diff
@@ -179,3 +182,91 @@ def test_update_stochastic_material():
     assert prop.distribution.name == "Normal"
     assert prop.distribution.mu == new_value
     assert prop.distribution.sigma == pytest.approx(sigma * new_value / mu)
+
+
+def test_get_property_unknown_relation_raises():
+    """Getting a property from an unknown material relation raises."""
+    material = Material.from_json(MATERIAL_LIB_DIR / "Ta6v.json")
+    with pytest.raises(ValueError, match="No material relation"):
+        material.get_property("UnknownRelation", "young_modulus")
```
MonPetitCanardArcEnCiel, line 192: Please write a test to check that `from_json` behaves as intended.
(except if this is checked elsewhere ?)

tests/material/test_material.py
```diff
@@ -179,3 +182,91 @@ def test_update_stochastic_material():
...
+    prop = MaterialProperty(
+        name="x",
+        value=1.0,
+        distribution=DistributionParameters(name="Triangular"),
+    )
+    material = Material(
+        name="m", material_relations=[MaterialRelation(name="r", properties=[prop])]
+    )
+    with pytest.raises(ValueError, match="Only deterministic, Normal and Uniform"):
+        material.update_from_dict({"x": 2.0})
```
MonPetitCanardArcEnCiel: Why not raise the error directly when setting the incorrect distribution in the `MaterialProperty` object ?

tests/tools/plotting/test_plot_model.py
```diff
@@ -51,3 +51,24 @@ def test_plot_model(tmp_wd, model, load_case):
...
+
+def test_plot_model_scalars(tmp_wd):
+    """Check that the scalar outputs of a model can be plotted in a scatter matrix.
+
+    A non-existing directory is passed to also check that it is created.
+    """
+    m = create_model("MockModel", "LC2")
+    m.EXTRA_INPUT_GRAMMAR_CHECK = True
+    m.execute()
+    figures = m.plot_results(directory_path="plots_subdir", data="SCALARS")
```
MonPetitCanardArcEnCiel: The dosctring states that the test should check that a non-existing directory is created, but there are no explicit checks that any files are created. Please check that the files are created. Also, what happen to those files ? Please check that they are cleaned at the end of the tests (is this what the `tmp_wd` argument is for ?).

tests/tools/plotting/test_plot_model.py
```diff
@@ -51,3 +51,24 @@ def test_plot_model(tmp_wd, model, load_case):
...
+    assert "scalars" in figures
```
MonPetitCanardArcEnCiel, line 68: Consider checking that the resulting plot looks like what it is supposed to, if this is not checked somewhere else.

tests/tools/test_base_composite_tool.py
```diff
@@ -0,0 +1,82 @@
...
+    [MyTool(), MyTool(working_directory="foo"), MyTool(root_directory="bar")],
+)
+@pytest.mark.parametrize(
+    ("working_directory_after_construction", "directory_naming_method"),
+    [
+        ("", None),
+        ("", DirectoryNamingMethod.NUMBERED),
+        (Path("my_dir"), None),
+    ],
+)
```
MonPetitCanardArcEnCiel, line 51: Why are there two `parameterize` decorators rather than one ? Is this intentional ?

tests/tools/test_base_composite_tool.py
```diff
@@ -0,0 +1,82 @@
...
+    tool.execute()
+    tool.save_results()
+
+    if directory_naming_method == DirectoryNamingMethod.NUMBERED:
+        assert tool.working_directory.name == "1"
+
+    assert (tool.working_directory / f"{expected_filename}_result.hdf5").is_file()
+    assert (
+        tool.working_directory / "MyTool" / f"{expected_subtool_filename}_result.hdf5"
+    ).is_file()
```
MonPetitCanardArcEnCiel, line 82: Consider checking that the content of the saved files is compliant, if this is not checked elsewhere.

Also same remark about remaining files at the end of the tests.

tests/tools/test_base_tool.py
```diff
@@ -0,0 +1,302 @@
...
+
+    with pytest.raises(
+        ValueError,
+        match=re.escape(
+            "You define keyword argument ``settings``. "
+            "If you want to define inputs, it must be done through the keyword argument "
+            "``inputs``."
+        ),
+    ):
+        tool.execute(settings=MySettings(foo="_"), bar="_")
```
MonPetitCanardArcEnCiel, line 89: I don't understand this error message. What happens here ?

From the text I understand that the problem is the `settings` keyword argument, but it is used elsewhere without error.

tests/tools/test_base_tool.py
```diff
@@ -0,0 +1,302 @@
...
+    ):
+        assert Path(tool.name).is_dir()
+
+    expected_filename = "MyTool"
+    if prefix == "":
+        assert (tool.working_directory / f"{expected_filename}_result.hdf5").is_file()
+    else:
+        assert (
+            tool.working_directory / f"{prefix}_{expected_filename}_result.hdf5"
+        ).is_file()
```
MonPetitCanardArcEnCiel, line 216: Please check that all generated files are cleaned at the end of the tests.

MonPetitCanardArcEnCiel, line 216: Consider checking that saved file content is compliant.

tests/tools/test_base_tool.py
```diff
@@ -0,0 +1,302 @@
...
+def test_save_and_load_results_pickle(tmp_wd):
+    """A pickled result can be saved and reloaded from disk."""
+    tool = MyTool()
+    tool.execute(foo="x")
+    tool.save_results(file_format="pickle")
+    path = tool.working_directory / "MyTool_result.pickle"
+    assert path.is_file()
+
+    loaded = BaseTool.load_results(path)
+    assert isinstance(loaded, BaseResult)
```
MonPetitCanardArcEnCiel, line 228: Consider checking that the content of the loaded object is compliant.

MonPetitCanardArcEnCiel (changes requested):

tests/material/test_material.py
SebastienBocquet: No, `tmp_wd` runs the test in a tmp dir (It is picked from gemseo)

tests/tools/test_base_composite_tool.py
SebastienBocquet: The serialization of the results to hdf5 is tested in `tests.tools.test_serialization`

tests/tools/test_base_tool.py
SebastienBocquet: It means that the user shall either pass settings and inputs as a single **kwargs, or use both `inputs` and `settings` args.
I added an explanation at the end of the error message. Please tell me if it is sufficient.

tests/tools/test_base_tool.py
SebastienBocquet: Should be the case thanks to `tmp_wd`

tests/tools/test_base_tool.py
SebastienBocquet: Done

tests/core/test_model_result.py
SebastienBocquet: I have set a field and now check the entire text.

SebastienBocquet: Done

tests/core/test_model_result.py
SebastienBocquet: Done

tests/material/test_material.py
SebastienBocquet: Done

tests/material/test_material.py
SebastienBocquet: It is checked on line 106 (`material_loaded = Material.from_json(file_name)...`)

tests/material/test_material.py
SebastienBocquet: Done

tests/tools/plotting/test_plot_model.py
SebastienBocquet: Done, I now check for the presence of the plot file. And there was a bug!
There is no need to clean thanks to `tmp_wd `.

tests/tools/test_base_tool.py
SebastienBocquet: The serialization of the results to hdf5 is tested in tests.tools.test_serialization

tests/tools/plotting/test_plot_model.py
SebastienBocquet: That is a strong choice whether or not testing the plots. We can discuss that IRL.

tests/tools/test_base_composite_tool.py
SebastienBocquet: It does a full factorial of the two parameterize.

tests/core/test_model_result.py
SebastienBocquet: I added a check on the variable names

src/vimseo/tools/base_tool.py
```diff
@@ -348,7 +348,7 @@ def _pre_process_options(self, **options):
                     msg = (
                         "You define keyword argument ``settings``. "
                         "If you want to define inputs, it must be done through "
-                        "the keyword argument ``inputs``."
+                        "the keyword argument ``inputs`` instead of passing `**kwargs`."
```
MonPetitCanardArcEnCiel:
```suggestion
             "``**kwargs`` is incompatible with keyword argument ``settings``. "
             "Either pass settings and inputs as a single `` **kwargs``, or use both ``inputs`` and ``settings`` args."
```

tests/tools/test_base_tool.py
MonPetitCanardArcEnCiel: I added a suggestion that I find clearer.

src/vimseo/tools/base_tool.py
SebastienBocquet: I have integrated your suggestion in a commit.

PR #62, opened 2026-07-16
fix(doc): silence matplotlib + plotly popups and fix doc build crash

MonPetitCanardArcEnCiel:
Closes #61 

## Problem
`tox run -e doc` pops real GUI windows / browser tabs during the doc build, and the build itself crashes (`mkdocs serve` exits 1). Four independent causes.

## Cause 1: matplotlib GUI windows
`[testenv:doc]` in `tox.ini` defined its own `set_env` block, which in tox >=4 replaces (rather than merges with) `[testenv]`'s `set_env` — silently dropping `MPLBACKEND=AGG`. matplotlib then fell back to an interactive backend and every example's `show=True` call opened a window.

**Fix**: extend the base `set_env` explicitly (`{[testenv]set_env}`).

## Cause 2: plotly browser tabs
Examples rendered via gemseo's plotly-backed `DatasetPlot`/`Curve.plot` (`file_format="html"`) call `Figure.show()`, whose default renderer outside a notebook is `"browser"` → `webbrowser.open()` — unrelated to matplotlib/`MPLBACKEND`.

**Fix**: force plotly's `"sphinx_gallery"` renderer via `PLOTLY_RENDERER`, add `ipython` as a doc dependency (required by that renderer).

## Cause 3 (fatal — this is why the build exits 1): griffe/docstring_inheritance crash
`src/vimseo/core/base_discipline_model.py`'s `BaseDisciplineModel` class has no docstring. The `docstring_inheritance.griffe` mkdocstrings extension, synthesizing one from the runtime-inherited `__doc__`, walks up to the containing module looking for an already-resolved docstring parser to reuse — the module has none, so it raises `RuntimeError: Cannot find a docstring for the parent of the object Class('BaseDisciplineModel', 32, 45)`, which is fatal to the whole `mkdocs serve` process.

**Fix**: add a docstring to BaseDisciplineModel, so griffe's own parsed docstring is used directly and no parser lookup is needed for this class.

## Cause 4 (non-fatal, but breaks example output): `plot_results` `AttributeError`
`base_integrated_model.py`'s `plot_results` only wraps the explicit-`directory_path` branch in `Path(...)`; the default (`self.archive_manager.job_directory`, a plain `str` `""` until a job directory has actually been created) is left unwrapped, so `directory_path.exists()` raises `AttributeError: 'str' object has no attribute 'exists'` for every example that calls `plot_results(show=True)` without an explicit `directory_path` — most of `01_models` and others.

**Fix**: wrap that branch in `Path(...)` too (`Path("")` already resolves to `Path(".")`, matching the documented `"defaults to cwd"` behavior).

## Verification
- `tox config -e doc` — resolved `set_env` includes `MPLBACKEND=AGG` and `PLOTLY_RENDERER=sphinx_gallery`.
- `tox exec -e doc -- python -c "import matplotlib as mpl; print(mpl.get_backend())"` — prints `AGG`.
- `tox exec -e doc -- python -c "import plotly.graph_objects as go; go.Figure().show()"` — completes without opening a browser tab.
- Full `tox run -e doc` build log: no `ERROR`/traceback, no more `AttributeError: 'str' object has no attribute 'exists'` warnings for the `01_models` examples. `Documentation built in 161.24 seconds`, server started clean.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -51,6 +51,7 @@ doc = [
     "black",
     "griffe-fieldz",
     "griffe-inherited-docstrings",
+    "ipython",
     "markdown-exec[ansi]",
     "mike",
     "mkdocs-bibtex",
--- a/requirements/doc.txt
+++ b/requirements/doc.txt
@@ -10,6 +10,8 @@ annotated-types==0.7.0
     # via pydantic
 anyio==4.13.0
     # via starlette
+asttokens==3.0.2
+    # via stack-data
 attrs==26.1.0
     # via
     #   jsonschema
@@ -62,6 +64,8 @@ databricks-sdk==0.106.0
     # via
     #   mlflow-skinny
     #   mlflow-tracing
+decorator==5.3.1
+    # via ipython
 dill==0.4.1
     # via openturns
 dnspython==2.8.0
@@ -78,6 +82,8 @@ et-xmlfile==2.0.0
     # via openpyxl
 eval-type-backport==0.2.2
     # via vimseo (pyproject.toml)
+executing==2.2.1
+    # via stack-data
 fastapi==0.136.1
     # via mlflow-skinny
 fastjsonschema==2.20.0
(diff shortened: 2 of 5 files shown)


SebastienBocquet:
No plot popping anymore, thanks! It was painful.

PR #64, opened 2026-07-17
Tecplot reader

MonPetitCanardArcEnCiel:
Adds a reader tool for an Ascii Tecplot file containing fields on a mesh.

Closes #63


src/vimseo/tools/io/reader_file_tecplot.py
```diff
@@ -0,0 +1,113 @@
...
+if TYPE_CHECKING:
+    from collections.abc import Mapping
+
+_FORMAT = "tecplot"
+
+
+def _rename_header_variables(text: str, variable_name_aliases: Mapping[str, str]) -> str:
+    """Rename the variables declared in a Tecplot ``VARIABLES`` header line.
+
+    meshio only recognizes coordinate variables named ``X``/``Y``/``Z``, while some
```
SebastienBocquet, line 46: I think that this docstring should remain very generic, not mentioning meshio related stuff.

src/vimseo/tools/io/reader_file_tecplot.py
```diff
@@ -0,0 +1,113 @@
...
+    return "".join(lines)
+
+
+class ReaderFileTecplotSettings(BaseFileReaderSettings):
+    """Settings of a Tecplot ASCII file reader."""
+
+    variable_name_aliases: dict[str, str] = Field(
+        default={},
+        description="The mapping from the variable names declared in the file's "
+        "VARIABLES header to the names expected by meshio's Tecplot reader "
```
SebastienBocquet: This docstring may only mention: "The mapping from the variable names declared in the file's "
        "VARIABLES header to the new names. Default value means no renaming."

src/vimseo/tools/io/reader_file_tecplot.py
```diff
@@ -0,0 +1,113 @@
...
+        file_name = options["file_name"]
+        directory_path = options["directory_path"]
+        if Path(file_name).suffix != self._EXTENSION:
+            msg = f"{self.__class__.__name__} requires the file suffix to be {self._EXTENSION}."
+            raise ValueError(msg)
+        file_path = (
+            file_name if directory_path == "" else Path(directory_path) / file_name
+        )
+
+        variable_name_aliases = options["variable_name_aliases"]
```
SebastienBocquet, line 148: Is it possible to do the renaming (necessary for meshio) automatically? Because it is mandatory as long as we use meshio as backend, and the user would not have to know the meshio convention. What the user could pass are the current names of the coordinates (as a tuple, new setting of ReaderFileTecplotSettings), and the internal logic handles the renaming for meshio. We could also raise an error if no X, Y or Z is found, since we know it will lead to meshio error? The variable_name_aliases may still be useful for renaming purpose.

src/vimseo/tools/io/reader_file_tecplot.py
```diff
@@ -0,0 +1,113 @@
...
+                line = line.replace(f'"{source_name}"', f'"{target_name}"')
+            lines[i] = line
+            break
+    return "".join(lines)
+
+
+class ReaderFileTecplotSettings(BaseFileReaderSettings):
+    """Settings of a Tecplot ASCII file reader."""
+
+    variable_name_aliases: dict[str, str] = Field(
```
SebastienBocquet, line 96: What do you think of: `old_to_new_names` ?

tests/tools/io/test_reader_file_tecplot.py
```diff
@@ -0,0 +1,71 @@
...
+0.0 1.0 1.0 0.0
+0.0 0.0 1.0 1.0
+0.0 0.0 0.0 0.0
+10.0 20.0 30.0 40.0
+1 2 3 4
+"""
+
+
+def test_reader_file_tecplot(tmp_wd):
+    file_name = "surface.dat"
```
SebastienBocquet, line 45: Can you add a docstring?

tests/tools/io/test_reader_file_tecplot.py
```diff
@@ -0,0 +1,71 @@
...
+    assert_allclose(field.mesh_points[:, 0], [0.0, 1.0, 1.0, 0.0])
+    assert len(field.mesh_cells) == 1
+    assert field.mesh_cells[0].type == "quad"
+    assert field.mesh_cells[0].data.shape == (1, 4)
+    assert field.point_variable_names == ["Pressure"]
+    assert_allclose(field.point_data["Pressure"], [10.0, 20.0, 30.0, 40.0])
+
+
+def test_reader_file_tecplot_wrong_extension(tmp_wd):
+    file_name = "surface.txt"
```
SebastienBocquet, line 90: Can you add a docstring?

src/vimseo/utilities/fields.py
```diff
@@ -46,12 +47,32 @@ def point_variable_names(self) -> Iterable[str]:
         return list(self.point_data.keys())
 
     @classmethod
-    def load(cls, path: Path | str):
-        field = read(path)
+    def from_mesh(cls, path: Path | str, mesh: Mesh) -> Field:
```
SebastienBocquet: It seems that the existing `load` method can do the job, no? It seems that you use the pattern `from_mesh(path, read(path))`, which is equivalent to `load(path)`.

MonPetitCanardArcEnCiel: Claude suggested this. I think the rationale was that, this way `load` may evolve to accomodate for other sources/methods without breaking the reader. What do you think ?

tests/tools/io/test_reader_file_tecplot.py
MonPetitCanardArcEnCiel: Done

tests/tools/io/test_reader_file_tecplot.py
MonPetitCanardArcEnCiel: Done

src/vimseo/tools/io/reader_file_tecplot.py
MonPetitCanardArcEnCiel: This is a protected method, not intended for public use, so I think it is defensible to have a more specific doctring that informs the developers on what problem this method is meant to adress

src/vimseo/utilities/fields.py
SebastienBocquet: Ok for `from_mesh`, but the signature needs to be changed to handle the case of a pure in-memory mesh: the path arg needs to be optional. So I propose: `def from_mesh(mesh: Mesh, path: Path | str = "")`

src/vimseo/tools/io/reader_file_tecplot.py
MonPetitCanardArcEnCiel: Done

src/vimseo/tools/io/reader_file_tecplot.py
MonPetitCanardArcEnCiel: I find it less clear. I updated the docstring to explain the order.

src/vimseo/tools/io/reader_file_tecplot.py
MonPetitCanardArcEnCiel: Done

SebastienBocquet:
I propose additional modifications (after exchanging with Claude), that are worth to do now before Field usage grows:

- Renaming Field => MeshField. Module name field.py can be left as is.

- FieldResult is composed of a `list[Field]` instead of a `Field`, since many Field usage will involve a series of Field. The default value would be an empty list.

- Field class. Below are several proposals/fixes: Wrong annotations for some fields, eq=False to avoid misleading field_1 == field_2 is True, better docstring,  a post_init to enforce string type for the path (otherwise the serialisation to hdf5 would have two different behaviours whether path is a Path or an str), 

```
@dataclass(eq=False)
class Field:
    """A field defined over a mesh.

    The field holds the mesh geometry together with the variables defined at its
    points and cells, as read by ``meshio``.
    """

    mesh_points: ndarray | None = None
    """The coordinates of the mesh points, of shape ``(n_points, 3)``."""

    mesh_cells: list[CellBlock] = field(default_factory=list)
    """The mesh cells, one block per cell type."""

    point_data: dict[str, ndarray] = field(default_factory=dict)
    """The variables defined at the mesh points."""

    cell_data: dict[str, list[ndarray]] = field(default_factory=dict)
    """The variables defined at the cells, one array per cell block."""

    path: str = ""
    """The path of the file the field was read from, empty if none."""

    def __post_init__(self) -> None:
        """Cast the path to a string, so that a ``Path`` is also accepted."""
        self.path = str(self.path)
```

- The two `properties cell_variable_names` and `point_variable_names` shall return `list[str]`

- Add test coverage focused on Field logic:

```
# tests/utilities/test_fields.py
@pytest.fixture
def unit_square_mesh() -> Mesh:
    """A two-triangle unit square with a linear pressure field."""
    return Mesh(
        points=[[0.0, 0.0], [1.0, 0.0], [1.0, 1.0], [0.0, 1.0]],
        cells=[("triangle", [[0, 1, 2], [0, 2, 3]])],
        point_data={"pressure": [0.0, 1.0, 2.0, 1.0]},
    )


def test_from_mesh(unit_square_mesh):
    """Check that a field is built from an already-read mesh."""
    field = Field.from_mesh(unit_square_mesh, "source.dat")
    assert field.path == "source.dat"
    assert field.point_variable_names == ["pressure"]
    assert_allclose(field.mesh_points, unit_square_mesh.points)


def test_from_mesh_without_path(unit_square_mesh):
    """Check that a field can be built from a mesh with no file provenance."""
    assert Field.from_mesh(unit_square_mesh).path == ""


def test_load(tmp_wd, unit_square_mesh):
    """Check that a field is loaded from a mesh file."""
    unit_square_mesh.write("mesh.vtk")
    field = Field.load("mesh.vtk")
    assert field.path == "mesh.vtk"
    assert_allclose(field.point_data["pressure"], [0.0, 1.0, 2.0, 1.0])
test_from_mesh_without_path is the one that pins the signature change — it doesn't compile against from_mesh(path, mesh), which makes the argument for you.
```

Two coverage gaps in the reader test while you're in there, unrelated to from_mesh:

the variable_name_aliases={} branch ([reader_file_tecplot.py:110-111](vscode-webview://1u34rpbho1bjlrua0hbd99r6p4f1i1sf5cff13eh1cl7copt7vf1/src/vimseo/tools/io/reader_file_tecplot.py#L110-L111)) is never exercised — a file already using X/Y/Z takes the other code path entirely;
directory_path != "" ([line 101-103](vscode-webview://1u34rpbho1bjlrua0hbd99r6p4f1i1sf5cff13eh1cl7copt7vf1/src/vimseo/tools/io/reader_file_tecplot.py#L101-L103)) is never exercised either.

src/vimseo/utilities/fields.py
MonPetitCanardArcEnCiel: Done

SebastienBocquet:
Your updates are OK for me. Please give me confirmation that you have finished.

MonPetitCanardArcEnCiel:
I'm finished provided the coverage is OK, as I did not see the updated Codecov report

Also I did not cover the unrelated uncovered lines identified by Claude, but this can be addressed in a dedicated issue.

SebastienBocquet:
I fixed the codecov upload. Ok to treat the missing coverage in another PR.

